PR #3, opened 2024-07-06
service/upower: add upower service

bdebiase:
Adds a UPower service to interact with system battery devices. I only implemented properties I thought would be useful, but more can be added later on.

This is my first shot at C++ so any suggestions would be greatly appreciated!


src/services/upower/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+set_source_files_properties(org.freedesktop.UPower.xml PROPERTIES
+	CLASSNAME DBusUPowerApp
```
outfoxxed: I'd change this to `DBusUPowerDaemon`, `DBusUPowerServer` or something like that.

src/services/upower/core.cpp
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	this->readyDevices.insertObject(device);
+	qCDebug(logUPower) << "UPowerDevice" << device->path() << "ready";
+}
+
+void UPower::onDeviceDestroyed(QObject* object) {
+	auto* device = static_cast<UPowerDevice*>(object); // NOLINT
+	this->readyDevices.removeObject(device);
+}
```
outfoxxed: Firstly add a debug log, secondly make sure mDisplayDevice is set to null if it was the one destroyed. Additionally remove it from `mDevices`.

src/services/upower/core.hpp
```diff
@@ -0,0 +1,51 @@
...
+namespace qs::service::upower {
+
+class UPower: public QObject {
+	Q_OBJECT;
+	QML_NAMED_ELEMENT(UPower);
+	QML_SINGLETON;
+	/// UPower's DisplayDevice for your system.
+	///
+	/// This is typically the device that is used for displaying information in desktop environments.
+	Q_PROPERTY(UPowerDevice* displayDevice READ displayDevice NOTIFY displayDeviceChanged);
```
outfoxxed: Add a note that it can be null

src/services/upower/device.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+bool UPowerDevice::isValid() const { return this->device->isValid(); }
+QString UPowerDevice::address() const { return this->device->service(); }
+QString UPowerDevice::path() const { return this->device->path(); }
+
+UPowerDeviceType::Enum UPowerDevice::type() const {
+	return static_cast<UPowerDeviceType::Enum>(this->pType.get());
+}
+bool UPowerDevice::powerSupply() const { return this->pPowerSupply.get(); }
```
outfoxxed: leave a newline

src/services/upower/device.cpp
```diff
@@ -0,0 +1,120 @@
...
+qreal UPowerDevice::energyFull() const { return this->pEnergyFull.get(); }
+qreal UPowerDevice::energyRate() const { return this->pEnergyRate.get(); }
+qlonglong UPowerDevice::timeToEmpty() const { return this->pTimeToEmpty.get(); }
+qlonglong UPowerDevice::timeToFull() const { return this->pTimeToFull.get(); }
+qreal UPowerDevice::percentage() const { return this->pPercentage.get(); }
+bool UPowerDevice::isPresent() const { return this->pIsPresent.get(); }
+UPowerDeviceState::Enum UPowerDevice::state() const {
+	return static_cast<UPowerDeviceState::Enum>(this->pState.get());
+}
+qreal UPowerDevice::capacity() const { return this->pCapacity.get(); }
```
outfoxxed: more newlines

src/services/upower/device.cpp
```diff
@@ -0,0 +1,120 @@
...
+QString UPowerDevice::iconName() const { return this->pIconName.get(); }
+
+void UPowerDevice::onGetAllFinished() { emit this->ready(); }
```
outfoxxed: Remove this and replace the connection with

```cpp
QObject::connect(&this->deviceProperties, &DBusPropertyGroup::getAllFinished, this, &UPowerDevice::ready);
```

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+	Q_ENUM(Enum);
+
+	Q_INVOKABLE static QString toString(UPowerDeviceType::Enum type);
+};
+
+///! A device exposed through the UPower system service.
+class UPowerDevice: public QObject {
+	Q_OBJECT;
+	// clang-format off
+	/// The type of power device.
```
outfoxxed: `power` is redundant

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+	///
+	/// > [!INFO] When checking if the device is a battery, also verify that `powerSupply` is set to `true`.
+	Q_PROPERTY(UPowerDeviceType::Enum type READ type NOTIFY typeChanged);
+	/// If the device is a power supply.
+	Q_PROPERTY(bool powerSupply READ powerSupply NOTIFY powerSupplyChanged);
+	/// Current energy level of the device in watt-hours.
+	Q_PROPERTY(qreal energy READ energy NOTIFY energyChanged);
+	/// Maximum energy capacity of the device in watt-hours
+	Q_PROPERTY(qreal energyFull READ energyFull NOTIFY energyFullChanged);
```
outfoxxed: Would call this `capacity`.

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+	/// Rate of energy change in watts (positive when charging, negative when discharging).
+	Q_PROPERTY(qreal energyRate READ energyRate NOTIFY energyRateChanged);
```
outfoxxed: Would call this `changeRate` or something like that.

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+public:
+	enum Enum {
+		Unknown = 0,
+		Charging = 1,
+		Discharging = 2,
+		Empty = 3,
+		FullyCharged = 4,
+		PendingCharge = 5,
+		PendingDischarge = 6,
+	};
```
outfoxxed: Comment PendingCharge and PendingDischarge. Assuming thats when the firmware decides to not charge/discharge until a condition is met but its unclear.

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+	/// Will be set to `0` if charging.
+	Q_PROPERTY(qreal timeToEmpty READ timeToEmpty NOTIFY timeToEmptyChanged);
+	/// Estimated time until the device is fully charged, in seconds.
+	///
+	/// Will be set to `0` if discharging.
+	Q_PROPERTY(qreal timeToFull READ timeToFull NOTIFY timeToFullChanged);
+	/// Current charge level as a percentage.
+	Q_PROPERTY(qreal percentage READ percentage NOTIFY percentageChanged);
+	/// If the device is currently present in the system.
+	Q_PROPERTY(bool isPresent READ isPresent NOTIFY isPresentChanged);
```
outfoxxed: This is unclear. Is this like if you have a removable battery and took it out or a charger that is unplugged? Add some more detail, and possibly rename to `connected` if that is the case.

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+	/// Current state of the device.
+	Q_PROPERTY(UPowerDeviceState::Enum state READ state NOTIFY stateChanged);
+	/// Capacity of the device as a percentage of its original capacity.
+	///
+	/// > [!WARNING] This property will be useless if the device comes from the system's [DisplayDevice](../upower).
+	Q_PROPERTY(qreal capacity READ capacity NOTIFY capacityChanged);
```
outfoxxed: Would this be for cases like battery degradation? Not sure capacity is the best name for that if so. Also clear up if `percentage` is `energy / energyFull` or `energy / energyFull * capacity`.

Additionally why the warning? Explain what conditions make the value useless so users can try to figure it out themselves on a non default device, and if possible use some non random value like `0` if the device does not support it, in addition to a `capacitySupported` property.

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+	/// Name of the icon representing the current state of the device.
+	Q_PROPERTY(QString iconName READ iconName NOTIFY iconNameChanged);
```
outfoxxed: Note the default value when none is given by upower.

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+///! A device exposed through the UPower system service.
+class UPowerDevice: public QObject {
+	Q_OBJECT;
+	// clang-format off
+	/// The type of power device.
+	///
+	/// > [!INFO] When checking if the device is a battery, also verify that `powerSupply` is set to `true`.
+	Q_PROPERTY(UPowerDeviceType::Enum type READ type NOTIFY typeChanged);
+	/// If the device is a power supply.
+	Q_PROPERTY(bool powerSupply READ powerSupply NOTIFY powerSupplyChanged);
```
outfoxxed: Is this redundant with `type` or does it span more than just `LinePower`?

src/services/upower/device.hpp
```diff
@@ -0,0 +1,174 @@
...
+	dbus::DBusProperty<double> pEnergy {this->deviceProperties, "Energy"};
+	dbus::DBusProperty<double> pEnergyFull {this->deviceProperties, "EnergyFull"};
+	dbus::DBusProperty<double> pEnergyRate {this->deviceProperties, "EnergyRate"};
+	dbus::DBusProperty<qlonglong> pTimeToEmpty {this->deviceProperties, "TimeToEmpty"};
+	dbus::DBusProperty<qlonglong> pTimeToFull {this->deviceProperties, "TimeToFull"};
+	dbus::DBusProperty<double> pPercentage {this->deviceProperties, "Percentage"};
+	dbus::DBusProperty<bool> pIsPresent {this->deviceProperties, "IsPresent"};
+	dbus::DBusProperty<unsigned int> pState {this->deviceProperties, "State"};
+	dbus::DBusProperty<double> pCapacity {this->deviceProperties, "Capacity"};
+	dbus::DBusProperty<QString> pIconName {this->deviceProperties, "IconName"};
```
outfoxxed: use q* types for numbers

src/services/upower/org.freedesktop.UPower.Device.xml
```diff
@@ -0,0 +1,17 @@
...
+		<property name="TimeToEmpty" type="x" access="read"/>
+		<property name="TimeToFull" type="x" access="read"/>
+		<property name="Percentage" type="d" access="read"/>
+		<property name="Temperature" type = "d" access="read"/>
+		<property name="IsPresent" type="b" access="read"/>
+        <property name="State" type="u" access="read"/>
+		<property name="Capacity" type = "d" access="read"/>
+		<property name="IconName" type="s" access="read"/>
+    </interface>
+</node>
```
outfoxxed: This has mixed tabs and spaces. Use all tabs.

src/services/upower/org.freedesktop.UPower.xml
```diff
@@ -0,0 +1,7 @@
+<node>
+    <interface name="org.freedesktop.UPower">
+		<method name="EnumerateDevices">
+			<arg direction="out" type="ao" name="devices"/>
+		</method>
+    </interface>
+</node>
```
outfoxxed: see above

src/services/upower/core.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include "../../core/model.hpp"
+#include "dbus_app.h"
+#include "device.hpp"
+
+namespace qs::service::upower {
+
+class UPower: public QObject {
+	Q_OBJECT;
+	QML_NAMED_ELEMENT(UPower);
+	QML_SINGLETON;
```
outfoxxed: Instead of using a singleton, use a global instance and a singleton as a proxy object as is done in [mpris](https://github.com/outfoxxed/quickshell/blob/master/src/services/mpris/watcher.hpp). This allows known state to persist across reloads instead of re-querying it every time.

outfoxxed (changes requested):
Thanks a lot!

Looks pretty good overall, just a few UAFs and nitpicks.

src/services/upower/org.freedesktop.UPower.Device.xml
```diff
@@ -0,0 +1,17 @@
...
+    <interface name="org.freedesktop.UPower.Device">
+		<property name="Type" type="u" access="read"/>
+		<property name="PowerSupply" type="b" access="read"/>
+		<property name="Energy" type="d" access="read"/>
+		<property name="EnergyFull" type="d" access="read"/>
+		<property name="EnergyRate" type="d" access="read"/>
+		<property name="TimeToEmpty" type="x" access="read"/>
+		<property name="TimeToFull" type="x" access="read"/>
+		<property name="Percentage" type="d" access="read"/>
+		<property name="Temperature" type = "d" access="read"/>
```
outfoxxed: Add a prop for this, its the only missing one as far as I can tell.

outfoxxed (changes requested):

src/services/upower/CMakeLists.txt
bdebiase: I agree, I'm going with DBusUPowerService as that may be more suitable.

src/services/upower/device.hpp
bdebiase: Changed it to `energyCapacity`.

src/services/upower/device.hpp
bdebiase: This property is relevant for batteries that can be removed from the device like batteries from older laptops. I think `isPresent` or `isPresentInBay` is still more suitable as it stays more true its behavior. I've changed the documentation and added a warning to be more specific. Let me know if there is a better way to address this. 
```cpp
/// If the power source is present in the bay or slot, useful for hot-removable batteries.
///
/// [!WARNING] If the device `type` is not `Battery`, then the property will be always be `false`.
Q_PROPERTY(bool isPresent READ isPresent NOTIFY isPresentChanged);
```

src/services/upower/device.hpp
bdebiase: I changed `capacity` to `healthPercentage` to make its behavior more clear. As for the warning, I don't think there is a *real* way to check if capacity it is supported so I can just set it to `false` for the DisplayDevice and `true` for any other device. The documentation for `percentage` has also been changed to clarify that it is equivalent to `energy / energyFull`.

src/services/upower/org.freedesktop.UPower.Device.xml
bdebiase: I removed this property since my device does not support it meaning I cannot test it to see if it works on DisplayDevice as well as normal devices. I am willing to add it back, but I don't see it as being necessary for a simple battery service.

src/services/upower/device.hpp
bdebiase: According to the DBus API reference, it's necessary to make sure both `powerSupply` is true and `type` is `Battery` for it to be considered a valid device. To avoid confusion, there are two ways I can think of making this simpler: 
1. Replacing it with an `isBattery` property:
   * This property will return true if both requirements are met (`powerSupply` is true and `type` is `Battery`).
   * This approach may introduce confusion if it coexists with the `type` property. To mitigate this, we might consider removing the `type` property, assuming the service is primarily used for batteries.
2. Masking it in the `type` property:
   * Modify the `type` property to return `Battery` only if `powerSupply` is also true. Otherwise, it will return `Unknown`.
   * This approach simplifies the interface by consolidating the logic into the `type` property, but introduces some hidden behavior.

I'm leaning more towards the first approach and removing the `type` property as it simplifies the interface while focusing more on battery devices, but I am very much open to any other ideas.

src/services/upower/device.hpp
outfoxxed: sounds good

src/services/upower/device.hpp
outfoxxed: nit: doesn't really need to be a warning, just normal text is fine

src/services/upower/device.hpp
outfoxxed: Looks like the property will be unset if it isn't supported, meaning you can check with `AbstractDBusProperty::exists()`.

src/services/upower/device.hpp
outfoxxed: Yeah thats a bit weird. maybe make an `isLaptopBattery` property that just checks both. I'd leave the `type` amd `powerSupply` properties though, while specifying powersupply means a power supply *for this computer*. Might be helpful to test if line power is classified this way.

src/services/upower/org.freedesktop.UPower.Device.xml
outfoxxed: yeah don't worry about it then

src/services/upower/core.cpp
```diff
@@ -107,4 +119,23 @@ void UPower::registerDevice(const QString& path) {
...
 
+UPower* UPower::instance() {
+	static UPower* instance = new UPower(); // NOLINT
+	return instance;
+}
+
+UPowerQml::UPowerQml(QObject* parent): QObject(parent) {
+	// clang-format off
+	QObject::connect(UPower::instance(), &UPower::displayDeviceChanged, this, &UPowerQml::displayDeviceChanged);
+	// clang-format on
```
outfoxxed: nit: leave single line connections formatted normally

src/services/upower/device.cpp
```diff
@@ -77,44 +76,48 @@ UPowerDevice::UPowerDevice(const QString& path, QObject* parent): QObject(parent
...
 	return static_cast<UPowerDeviceState::Enum>(this->pState.get());
 }
-qreal UPowerDevice::capacity() const { return this->pCapacity.get(); }
-QString UPowerDevice::iconName() const { return this->pIconName.get(); }
 
-void UPowerDevice::onGetAllFinished() { emit this->ready(); }
+qreal UPowerDevice::healthPercentage() const { return this->pHealthPercentage.get(); }
+bool UPowerDevice::healthSupported() const { return this->path() != "/org/freedesktop/UPower/devices/DisplayDevice"; }
+QString UPowerDevice::iconName() const { return this->pIconName.get(); }
+// clang-format on
```
outfoxxed: nit: just let the formatter do what it wants here while keeping the line spacing correct

src/services/upower/device.hpp
outfoxxed: Checked my power adapter with dbus and it seems to confirm this. `powerSupply` means "is supplying power to this computer"

src/services/upower/device.hpp
bdebiase: Didn't know this, thanks!

src/services/upower/device.hpp
bdebiase: Seems good!

src/services/upower/device.hpp
bdebiase: Turns out the `exists()` method always comes out to true; even if it is a DisplayDevice which just returns `0`. My guess is UPower has all of its properties visible regardless of whether or not the device supports it.

The current way I implemented it is kinda arbitrary (by checking the device path) but that's all I can think of for implementing the `healthSupported` property - other than just checking if it is already `0`.

outfoxxed: I think its fair to say capacity isn't supported if its zero. better than checking if it's a display device at least.

src/services/upower/device.cpp
```diff
@@ -0,0 +1,129 @@
...
+	switch (status) {
+	case UPowerDeviceState::Unknown: return "Unknown";
+	case UPowerDeviceState::Charging: return "Charging";
+	case UPowerDeviceState::Discharging: return "Discharging";
+	case UPowerDeviceState::Empty: return "Empty";
+	case UPowerDeviceState::FullyCharged: return "Fully Charged";
+	case UPowerDeviceState::PendingCharge: return "Pending Charge";
+	case UPowerDeviceState::PendingDischarge: return "Pending Discharge";
+	}
+}
```
outfoxxed: Add a default case. QML just passes in the integer representation which can be invalid and run off the end of the function function which is UB.

src/services/upower/device.cpp
```diff
@@ -0,0 +1,129 @@
...
+	case UPowerDeviceType::Video: return "Video";
+	case UPowerDeviceType::OtherAudio: return "Other Audio";
+	case UPowerDeviceType::RemoteControl: return "Remote Control";
+	case UPowerDeviceType::Printer: return "Printer";
+	case UPowerDeviceType::Scanner: return "Scanner";
+	case UPowerDeviceType::Camera: return "Camera";
+	case UPowerDeviceType::Wearable: return "Wearable";
+	case UPowerDeviceType::Toy: return "Toy";
+	case UPowerDeviceType::BluetoothGeneric: return "Bluetooth Generic";
+	}
```
outfoxxed: same as above

src/services/upower/core.hpp
```diff
@@ -0,0 +1,66 @@
...
+};
+
+class UPowerQml: public QObject {
+	Q_OBJECT;
+	QML_NAMED_ELEMENT(UPower);
+	QML_SINGLETON;
+	/// UPower's DisplayDevice for your system.
+	///
+	/// This is typically the device that is used for displaying information in desktop environments. Can be `null`.
+	Q_PROPERTY(UPowerDevice* displayDevice READ displayDevice NOTIFY displayDeviceChanged);
```
outfoxxed: Mention that this is an aggregate device and not a physical device, and therefore is not present in the device list.

src/services/upower/core.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <qqmllist.h>
+#include <qtmetamacros.h>
+
+#include "../../core/model.hpp"
+#include "dbus_service.h"
+#include "device.hpp"
+
+namespace qs::service::upower {
+
+class UPower: public QObject {
```
outfoxxed, line 17: Would be nice to have a property exposed for `OnBattery`.

src/services/upower/device.hpp
```diff
@@ -0,0 +1,182 @@
...
+		Toy = 27,
+		BluetoothGeneric = 28,
+	};
+	Q_ENUM(Enum);
+
+	Q_INVOKABLE static QString toString(UPowerDeviceType::Enum type);
+};
+
+///! A device exposed through the UPower system service.
+class UPowerDevice: public QObject {
```
outfoxxed, line 80: Would be nice to expose `NativePath`.

outfoxxed (changes requested):
Noticed a few more things while testing.

Here's my tester if you don't have one, but I assume you do

<details>

```qml
import QtQuick
import QtQuick.Controls
import QtQuick.Layouts
import Quickshell
import Quickshell.Services.UPower

ShellRoot {
	Component {
		id: comp

		Rectangle {
			required property var modelData;

			implicitWidth: layout.implicitWidth + 10
			implicitHeight: layout.implicitHeight + 10
			color: palette.active.base

			ColumnLayout {
				id: layout

				Label { text: `is display device: ${modelData == UPower.displayDevice}` }
				Label { text: `device type: ${UPowerDeviceType.toString(modelData.type)}` }
				Label { text: `is power supply: ${modelData.powerSupply}` }
				Label { text: `energy: ${modelData.energy}` }
				Label { text: `energy capacity: ${modelData.energy}` }
				Label { text: `change rate: ${modelData.changeRate}` }
				Label { text: `time to empty: ${modelData.timeToEmpty}` }
				Label { text: `time to full: ${modelData.timeToFull}` }
				Label { text: `percentage: ${modelData.percentage}` }
				Label { text: `is present: ${modelData.isPresent}` }
				Label { text: `state: ${UPowerDeviceState.toString(modelData.state)}` }
				Label { text: `health percentage: ${modelData.healthPercentage}` }
				Label { text: `health supported: ${modelData.healthSupported}` }
				Label { text: `icon name: ${modelData.iconName}` }
				Label { text: `is laptop battery: ${modelData.isLaptopBattery}` }
			}
		}
	}

	FloatingWindow {
		color: contentItem.palette.active.window

		ListView {
			anchors.fill: parent
			spacing: 10

			model: UPower.devices
			delegate: comp
		}

		Loader {
			anchors.right: parent.right
			active: UPower.displayDevice != null

			sourceComponent: BoundComponent {
				sourceComponent: comp
				property var modelData: UPower.displayDevice
			}
		}
	}
}
```

</details>

src/services/upower/device.hpp
```diff
@@ -117,6 +117,10 @@ class UPowerDevice: public QObject {
 	///
 	/// This will be equivalent to `type == Battery && powerSupply == true`.
  	Q_PROPERTY(bool isLaptopBattery READ isLaptopBattery NOTIFY isLaptopBatteryChanged);
+	/// Native path of the device specific to your OS.
+	Q_PROPERTY(QString nativePath READ nativePath NOTIFY nativePathChanged);
+	/// If the device is currently running on battery power, or discharging.
+	Q_PROPERTY(bool onBattery READ onBattery NOTIFY onBatteryChanged);
```
outfoxxed: OnBattery is a global property, not a device property, see https://upower.freedesktop.org/docs/UPower.html

outfoxxed (changes requested):
After this please squash+rebase and it should be good to merge.

src/services/upower/device.hpp
bdebiase: I apologize, that completely slipped by me!

outfoxxed:
LGTM, thanks!

bdebiase:
Thanks for everything!

PR #8, opened 2024-11-02
Add i3/Sway support

nydragon:
This pr adds support for retrieving data from Sway's IPC.

Features:

- [x] Get last IPC object
- [x] Dispatch Sway commands
- [x] Workspaces
   - [x] Retrieve all
   - [x] Listen to data changes via signals
   - [x] Force full refresh of workspaces
   - [x] Get active 
   - [x] Live update
- [x] Monitors
   - [x] Retrieve all
   - [x] Listen to data changes via signals
   - [x] Force full refresh of monitors
   - [x] Get active
   - [x] Live update
   
Additional Todo:
- [x] documentation


outfoxxed:
Looks largely copied from hyprland ipc at this point so I figured I'd mention a few things:
- Make sure you only copy what makes sense to copy, don't keep what doesn't fit.
- If you can leave the dispatch socket open you should. You can't in hyprland, idk about i3/sway
- If you aren't implementing any sway specific features, this should work with i3 and be named as such.

nydragon:
Thanks for the feedback, I wanted to stay mainly in line with the existing code, but I do have some questions:

1. Sway calls monitors "outputs", should that be reflected in Quickshell or is it better to be have uniform naming across different compositors ? Or "dispatch" is called "run commands"
2. Should I move the x11 and wayland folders into a compositor folder and create a `shared` folder in there? Or where would I put the sway-i3 code ?
3. We could abstract a part of the logic, namely the workspace/monitors refresh functions, across Hyprland & Sway/i3, does that make sense ?

outfoxxed:
1. Monitor. Dispatch under hyprland is only for commands with no response as we don't currently have a promise type.
2. For now src/x11/i3/ipc is fine. Hyprland is under the wayland folder because it has to use cmake functions defined in the wayland folder.
3. Yes. This should happen, but not in this PR.

Also make sure you have a way to turn a sway monitor into a qs monitor.

outfoxxed:
You've force pushed a lot, this ready for review?

nydragon:
Yep, you could review it,

I would still want to implement more i3/Sway IPC features, but I think that will be for a later PR

CMakeLists.txt
```diff
@@ -55,6 +55,8 @@ boption(HYPRLAND "  Hyprland" ON REQUIRES WAYLAND)
 boption(HYPRLAND_IPC "    Hyprland IPC" ON REQUIRES HYPRLAND)
 boption(HYPRLAND_GLOBAL_SHORTCUTS "    Hyprland Global Shortcuts" ON REQUIRES HYPRLAND)
 boption(HYPRLAND_FOCUS_GRAB "    Hyprland Focus Grabbing" ON REQUIRES HYPRLAND)
+boption(I3 "  I3/Sway" ON)
+boption(I3_IPC "    I3/Sway IPC" ON REQUIRES I3)
```
outfoxxed, line 59: Add a note in BUILD.md

src/x11/i3/ipc/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+	connection.cpp
+	qml.cpp
+	workspace.cpp
+	monitor.cpp
+)
+
+qt_add_qml_module(quickshell-i3-ipc
+	URI Quickshell.I3._Ipc
+    VERSION 0.1
+    DEPENDENCIES QtQml Quickshell
```
outfoxxed: Remove Quickshell from DEPENDENCIES. add_module_deps_light covers that.

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+#include <qtypes.h>
+
+#include "../../../core/model.hpp"
+#include "../../../core/qmlscreen.hpp"
+#include "connection.hpp"
+#include "monitor.hpp"
+#include "workspace.hpp"
+
+Q_LOGGING_CATEGORY(logi3Ipc, "quickshell.I3.ipc", QtWarningMsg);
+Q_LOGGING_CATEGORY(logi3IpcEvents, "quickshell.I3.ipc.events", QtWarningMsg);
```
outfoxxed: lowercase

src/x11/i3/ipc/connection.hpp
```diff
@@ -0,0 +1,133 @@
...
+	explicit I3Ipc();
+
+	void onEvent(I3IpcEvent* event);
+
+	void handleWorkspaceEvent(I3IpcEvent* event);
+
+	static QVector<std::tuple<EventCode, QJsonDocument>> parseResponse(QByteArray rawEvent);
+
+	QLocalSocket liveEventSocket;
+	QLocalSocket commandSocket;
```
outfoxxed: do we need 2?

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+QByteArray I3Ipc::buildRequestMessage(EventCode cmd, const QByteArray& payload) {
+	int len = static_cast<int>(payload.length());
+	QByteArray lenBytes = QByteArray(4, Qt::Initialization::Uninitialized);
+	QByteArray typeBytes = QByteArray(4, Qt::Initialization::Uninitialized);
+
+	std::memcpy(lenBytes.data(), &len, 4);
+	std::memcpy(typeBytes.data(), &cmd, 4);
+
+	QByteArray data = QByteArray(MAGIC.data()) + lenBytes + typeBytes + payload;
```
outfoxxed: Consider using std::bit_cast or QDataStream

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+	return data;
+}
+
+I3Ipc::I3Ipc() {
+	auto sock = qEnvironmentVariable("I3SOCK");
+
+	if (sock.isEmpty()) {
+		qWarning() << "$I3SOCK is unset. Trying $SWAYSOCK Cannot connect to I3.";
```
outfoxxed: grammar

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+QVector<Event> I3Ipc::makeRequest(const QByteArray& request) {
+	this->commandSocket.write(request);
+	this->commandSocket.flush();
+
+	if (this->commandSocket.waitForReadyRead()) {
+		auto dataRaw = this->commandSocket.readAll();
+		return I3Ipc::parseResponse(dataRaw);
+	} else {
+		qWarning(logi3Ipc) << "Timeout while waiting for read for" << request;
```
outfoxxed: qWarning -> qCWarning everywhere

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+	}
+
+	auto [type, data] = res.first();
+
+	auto workspaces = data.array();
+
+	const auto& mList = this->mWorkspaces.valueList();
+	auto names = QVector<QString>();
+
+	qInfo(logi3Ipc) << "There are" << workspaces.toVariantList().length() << "workspaces";
```
outfoxxed: qinfo -> qcinfo and this looks debug worthy at most

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+	auto removedWorkspaces = QVector<I3Workspace*>();
+
+	for (auto* workspace: mList) {
+		if (!names.contains(workspace->name())) {
+			removedWorkspaces.push_back(workspace);
+		}
+	}
+
+	qInfo(logi3Ipc) << "Removing" << removedWorkspaces.length() << "deleted workspaces.";
```
outfoxxed: debug

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+		return;
+	}
+
+	auto [type, data] = res.first();
+
+	auto monitors = data.array();
+	const auto& mList = this->mMonitors.valueList();
+	auto names = QVector<QString>();
+
+	qInfo(logi3Ipc) << "There are" << monitors.toVariantList().length() << "monitors";
```
outfoxxed: same as above

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+}
+
+I3Monitor* I3Ipc::monitorFor(QuickshellScreenInfo* screen) {
+	if (screen == nullptr) return nullptr;
+
+	return this->findMonitorByName(screen->name());
+}
+
+I3Workspace* I3Ipc::findWorkspaceByName(const QString& name) {
+	auto mList = this->mWorkspaces.valueList();
```
outfoxxed: why mList? `m` conventionally means "member".

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+ObjectModel<I3Monitor>* I3Ipc::monitors() { return &this->mMonitors; }
+ObjectModel<I3Workspace>* I3Ipc::workspaces() { return &this->mWorkspaces; }
+
+QString I3IpcEvent::type() const { return I3IpcEvent::eventToString(this->mCode); }
+QString I3IpcEvent::data() const { return QString::fromUtf8(this->mData.toJson()); }
+
+EventCode I3IpcEvent::intToEvent(uint32_t raw) {
+	if ((EventCode::WORKSPACE <= raw && raw <= EventCode::OUTPUT)
+	    || (EventCode::RUN_COMMAND <= raw && raw <= EventCode::GET_TREE))
```
outfoxxed: use all == unless i'm missing something

src/x11/i3/ipc/connection.hpp
```diff
@@ -0,0 +1,133 @@
...
+namespace qs::i3::ipc {
+
+constexpr std::string MAGIC = "i3-ipc";
+
+enum EventCode {
+	RUN_COMMAND = 0,    // NOLINT
+	GET_WORKSPACES = 1, // NOLINT
+	SUBSCRIBE = 2,
+	GET_OUTPUTS = 3, // NOLINT
+	GET_TREE = 4,    // NOLINT
```
outfoxxed: what lint is failing?

src/x11/i3/ipc/qml.hpp
```diff
@@ -0,0 +1,73 @@
...
+	QSDOC_TYPE_OVERRIDE(ObjectModel<qs::i3::ipc::I3Workspace>*);
+	Q_PROPERTY(UntypedObjectModel* workspaces READ workspaces CONSTANT);
+	// clang-format on
+	QML_NAMED_ELEMENT(I3);
+	QML_SINGLETON;
+
+public:
+	explicit I3IpcQml();
+
+	/// Execute an I3/Sway command [man 5 sway](https://man.archlinux.org/man/sway.5.en#COMMANDS)
```
outfoxxed: link https://i3wm.org/docs/userguide.html#list_of_commands instead probably

outfoxxed (changes requested):
Bunch of nitpicks. I assume you've gone through and tested everything manually as well, run the lints, etc. If you have a tester file send that along as well and I'll give it a try before merging.

src/x11/i3/ipc/connection.cpp
nydragon: Thought it'd be more concise than matching all 8 enum members

src/x11/i3/ipc/connection.hpp
nydragon: It doesn't like screaming snake case

src/x11/i3/ipc/connection.hpp
nydragon: Didn't find a way to circumvent a race condition coming from the live ipc event sending a message while still waiting for the command response, since I am returning the reply to the QS user

nydragon:
I have just noticed a crash related to those changes, don't know how to replicate it yet but I would need to fix that before this PR can progress

edit: fixed, was a lot more straightforward than I thought

outfoxxed:
Consider using QDataStream transactions instead. As is, if you break halfway through an event then the following event will be invalid and so on. check ipc.cpp for an example

src/x11/i3/ipc/connection.hpp
outfoxxed: ah yeah. use UpperCamelCase with an enum class.

src/x11/i3/ipc/connection.cpp
outfoxxed: this doesn't look like it matches all the enum members as is

src/x11/i3/ipc/connection.hpp
outfoxxed: If you're returning all events and not just requested ones you can use a single socket since you'd be racing anyway. Otherwise 2 is fine

src/x11/i3/ipc/connection.cpp
nydragon: It should be matching all except for the Unknown member

src/x11/i3/ipc/connection.hpp
nydragon: I am emitting all events through rawEvent, but dispatch returns to the user the IPC's reply to the requested command

nydragon:
> Consider using QDataStream transactions instead. As is, if you break halfway through an event then the following event will be invalid and so on. check ipc.cpp for an example

I ended up doing that with QByteArrays instead, if I encounter a broken message i try to find the next occurrence of the magic sequence and skip to there. However this is under the assumption that the Sway IPC does not write partial messages to the socket, and it does not seem to do so.

outfoxxed:
Do we ever encounter broken messages? i3 shouldn't be sending those at all. I think the best solution to read the header would be a QDataStream transaction, which lets you do a partial read pretty easily.

Basically:
```cpp
stream.startTransaction()
stream.startTransaction()
stream >> magic >> header bits >> size;
if (!stream.commitTransaction()) return;
stream >> data; // assuming you need size to find data. nested commitTransaction works as a barrier for the integrity of values read previosuly
if (!stream.commitTransaction()) return;
```
You can nest transactions and it works pretty well. This should be more reliable than skipping events. Failing commits roll back the stream, and if there's enough data on the next readyread it should work.

nydragon:
Alrighty, that should be implemented now, here is also a file to test the IPC impl:

<details><summary>Details</summary>
<p>

```qml
import QtQuick
import Quickshell
import Quickshell.I3

ShellRoot {
    PanelWindow {
        id: panel
        Timer {
            id: timer
        }
        Component.onCompleted: {
            const program = "foot"; // change to an installed program
            const longLog = false; // toggle logging of workspace & monitor objects
            const stringify = v => JSON.stringify(v, null, 2);
            console.log("Connecting to", I3.socketPath);
            I3.rawEvent.connect(e => {
                console.log("Event:", e.type);
            });
            I3.focusedWorkspaceChanged.connect(() => {
                console.log("Changed to workspace", I3.focusedWorkspace.name, I3.focusedMonitor.focusedWorkspace.name);
                console.log("Focused monitor's focused workspace and actual focused workspace should be the same:");
                console.log(`${I3.focusedMonitor.focusedWorkspace.name} and ${I3.focusedWorkspace.name}`);
            });
            I3.focusedMonitorChanged.connect(() => {
                console.log("Changed to monitor", I3.focusedMonitor.name);
            });
            for (const monitor of I3.monitors.values) {
                console.log("Monitor:", monitor.name);
                longLog && console.log(stringify(monitor));
            }
            for (const workspace of I3.workspaces.values) {
                console.log("Workspace:", workspace.name);
                longLog && console.log(stringify(workspace));
                I3.dispatch(`workspace ${workspace.num}`);
            }
            I3.dispatch(`workspace 13`);
            console.log(`Changing workspace response: ${stringify(I3.dispatch(`exec ${program}`))}`);
            console.log(`Bad request response: ${stringify(I3.dispatch(`quickshell`))}`);
            I3.dispatch(`workspace 1`);
            I3.refreshWorkspaces();
            I3.refreshMonitors();
            console.log(stringify(I3.monitorFor(panel.screen)));
        }
    }
}
```

</p>
</details>

outfoxxed:
Until you commit the transaction you shouldn't access data that is part of it, as it might not have been read. Also, rolling back the transaction to read again when you hit something invalid will just hang. On my phone rn, will do a proper code review again later.

src/x11/i3/ipc/connection.cpp
outfoxxed: Had to think about the expression for a minute for it to click but I see the intention now, mb.

src/x11/i3/ipc/connection.hpp
outfoxxed: Sorry, haven't had time to go over this in detail before now, will comment on the larger issue in main review.

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+	this->commandSocket.write(request);
+	this->commandSocket.flush();
+
+	if (this->commandSocket.waitForReadyRead()) {
+		return I3Ipc::parseResponse(this->commandSocketDs);
+	} else {
+		qCWarning(logI3Ipc) << "Timeout while waiting for read for" << request;
+		return {};
+	}
+}
```
outfoxxed: This is a problem I missed the first time. Blocking on the main thread is almost never acceptable in quickshell, because not stalling the UI is extremely important. Especially important is to never block the main thread in a way that depends on an arbitrary response from another program. This is why the hyprland IPC makeRequest is asynchronous with a callback.

It does make things harder than they have to be, but I consider it very important to never leave the main thread blocked on IO.

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+
+QJsonArray I3Ipc::dispatch(const QString& payload) {
```
outfoxxed: Similarly to the above, we can't return the response here without blocking. We need some kind of promise mechanism, but it's non-trivial to get right. I'd be happy to get input on it via the matrix channel or something if you have any, but don't try to deal with it in this PR.

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+			monitor->setFocusedWorkspace(this->mFocusedWorkspace);
+		}
+
+		QObject::connect(workspace, &QObject::destroyed, this, &I3Ipc::onFocusedWorkspaceDestroyed);
+		workspace->setFocus(true);
+		this->setFocusedMonitor(workspace->monitor());
+	}
+
+	emit this->focusedWorkspaceChanged();
+}
```
outfoxxed, line 208: I was under the impression that workspace focus is per-monitor in i3, in which case this should be a function on I3Monitor instead. Does i3 allow moving a workspace to another monitor by setting it as the focused one?

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+	// clang-format on
+
+	auto liveIPCconnectedCallback = [this]() {
+		auto payload = QByteArray(R"(["workspace","output"])");
+		auto message = I3Ipc::buildRequestMessage(EventCode::SUBSCRIBE, payload);
+
+		this->liveEventSocket.write(message);
+		this->liveEventSocket.flush();
+		emit this->connected();
+	};
```
outfoxxed: given this, do we need refreshMonitors/refreshWorkspaces to handle responses, or can it just use the same handler as any other events?

Also please move it to a function to match the code style. I'll try to make that clearer in the future.

src/x11/i3/ipc/connection.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+	void handleWorkspaceEvent(I3IpcEvent* event);
+
+	static QVector<std::tuple<EventCode, QJsonDocument>> parseResponse(QDataStream& rawEvent);
+
+	QLocalSocket liveEventSocket;
+	QDataStream liveEventSocketDs;
+
+	QLocalSocket commandSocket;
+	QDataStream commandSocketDs;
```
outfoxxed: given that we can't make requests synchronously until promises are implemented, can we remove this?

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+	for (auto& [type, data]: I3Ipc::parseResponse(this->liveEventSocketDs)) {
+		this->event.mCode = type;
+		this->event.mData = data;
+
+		this->onEvent(&this->event);
+		emit this->rawEvent(&this->event);
+	}
+}
+
+QVector<Event> I3Ipc::parseResponse(QDataStream& rawEvent) {
```
outfoxxed: `rawEvent` is a really weird name for this. Try `eventStream` or something that signifies its an input stream and not a singular event.

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+	QVector<std::tuple<EventCode, QJsonDocument>> events;
+	const int magicLen = 6;
+
+	while (!rawEvent.atEnd()) {
+		rawEvent.startTransaction();
+		rawEvent.startTransaction();
+
+		QByteArray buffer(MAGIC.length(), 0);
```
outfoxxed: Use std::array instead of a heap allocated, dynamically sized buffer.

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+		QByteArray payload(size, Qt::Uninitialized);
+
+		rawEvent.readRawData(payload.data(), size);
+
+		if (!rawEvent.commitTransaction()) break;
+
+		if (buffer != MAGIC.data()) {
+			qCWarning(logI3Ipc) << "No magic sequence found in string." << buffer;
+			break;
+		};
```
outfoxxed: I think receiving a bad magic sequence isn't a recoverable error. I would tear down the connection and re-create it, or just print an error in the log and stop reading, because something is definitely wrong. Breaking at least will not recover this.

outfoxxed:
While reviewing, I was going to test it and it crashed instantly under hyprland. It shouldn't work but it shouldn't crash either.
image

src/x11/i3/ipc/connection.cpp
```diff
@@ -0,0 +1,550 @@
...
+		if (workspace == nullptr) {
+			workspace = new I3Workspace(this);
+		}
+
+		if (workspaceData.isObject()) {
+			workspace->updateFromObject(workspaceData.toObject().toVariantMap());
+		}
+
+		this->mWorkspaces.insertObject(workspace);
+		qCInfo(logI3Ipc) << "Added workspace to list";
```
outfoxxed: If you're going to log things, make it detailed enough to be useful.

src/x11/i3/ipc/monitor.hpp
```diff
@@ -0,0 +1,100 @@
...
+	[[nodiscard]] qint32 width() const;
+	[[nodiscard]] qint32 height() const;
+	[[nodiscard]] qreal scale() const;
+	[[nodiscard]] bool focused() const;
+	[[nodiscard]] QVariantMap lastIpcObject() const;
+
+	void updateFromObject(const QVariantMap& obj);
+
+	void setFocusedWorkspace(I3Workspace* workspace);
+	void setFocus(bool focus);
```
outfoxxed, line 68: is there a reason we can't set these from qml via props or invocation?

outfoxxed (changes requested):
Might've missed things again, but this is what I saw this time. Apologies for the repeated re-reviews but I want to be sure quickshell's integrations are stable.

src/x11/i3/ipc/connection.cpp
nydragon: Alright, means I will just drop the other socket and remove the return value

src/x11/i3/ipc/connection.cpp
nydragon: There can only ever be one focused workspace at all times (i.e only one workspace has focused = true), they do also return whether a workspace is visible or not, and if a workspace is visible, it is the one present in the current_workspace field you get from `get_outputs`

src/x11/i3/ipc/connection.cpp
nydragon: Monitor events do not return any data, at least on Sway, there we'd need to fetch all outputs:

<details><summary>man</summary>
<p>

   0x80000001. OUTPUT
       Sent whenever an output is added, removed, or its configuration is changed.  The
       event  is a single object with the property change, which is a string containing
       the reason for the change. Currently, the only value for change is unspecified.

       Example Event:
           {
                "change": "unspecified"
           }


</p>
</details> 

But I am unsure about what exactly you mean by this message, events are generally handled in onEvent, and considering the previous comments, I will probably move the handling of get_outputs/get_workspaces to there and refresh* will only contain the request

src/x11/i3/ipc/monitor.hpp
nydragon: I did not intend to allow changing focused monitor that way, that's possible through the dispatch function for now, but it could definitely make sense

src/x11/i3/ipc/connection.cpp
outfoxxed: alright, not a problem.

src/x11/i3/ipc/connection.cpp
outfoxxed: That sounds good to me.

src/x11/i3/ipc/monitor.hpp
outfoxxed: dw about it for now.

nydragon:
Should be all fixed now

src/x11/i3/ipc/connection.cpp
```diff
@@ -137,44 +102,51 @@ void I3Ipc::eventSocketReady() {
...
+		if (strncmp(buffer.data(), MAGIC.data(), 6) != 0) {
+			qCWarning(logI3Ipc) << "No magic sequence found in string.";
+			this->reconnectIPC();
 			break;
 		};
 
-		if (I3IpcEvent::intToEvent(type) == EventCode::UNKNOWN) {
+		if (I3IpcEvent::intToEvent(type) == EventCode::Unknown) {
 			qCWarning(logI3Ipc) << "Received unknown event";
+			this->reconnectIPC();
```
outfoxxed: Don't need to reconnect for this one, as long as the event stream is still well-formed.

src/x11/i3/ipc/connection.cpp
```diff
@@ -390,18 +354,20 @@ void I3Ipc::refreshMonitors() {
 		this->mMonitors.removeObject(monitor);
 		delete monitor;
 	}
-}
+};
```
outfoxxed: ;

outfoxxed (changes requested):
Didn't see anything else at a glance, will give it a more thorough lookover again shortly.

outfoxxed:
We shouldn't be trying to write to the socket unless we're actually connected. I think hyprland IPC also does this though, now that I think about it. Will fix that later.
```
build/src/quickshell -p testi3.qml
  INFO: Launching config: "/home/admin/programming/outfoxxed/quickshell/i3/testi3.qml"
  INFO: Shell ID: "c4c77744f223921c7870881fbb46b0a8" Path ID "c4c77744f223921c7870881fbb46b0a8"
  INFO: Saving logs to "/run/user/1000/quickshell/by-id/jij3guvdns/log.log"
  WARN quickshell.I3.ipc: $I3SOCK is unset. Trying $SWAYSOCK.
  WARN quickshell.I3.ipc: $SWAYSOCK and I3SOCK are unset. Cannot connect to socket.
 DEBUG qml: Connecting to
  WARN: QIODevice::write (QLocalSocket): device not open
  WARN: QIODevice::write (QLocalSocket): device not open
 DEBUG qml: Changing workspace response: undefined
  WARN: QIODevice::write (QLocalSocket): device not open
 DEBUG qml: Bad request response: undefined
  WARN: QIODevice::write (QLocalSocket): device not open
  WARN: QIODevice::write (QLocalSocket): device not open
  WARN: QIODevice::write (QLocalSocket): device not open
 DEBUG qml: null
  INFO: Configuration Loaded
  ```

File: src/x11/i3/ipc/CMakeLists.txt [hunk 27]
```diff
@@ -0,0 +1,22 @@
...
+	qml.cpp
+	workspace.cpp
+	monitor.cpp
+)
+
+qt_add_qml_module(quickshell-i3-ipc
+	URI Quickshell.I3._Ipc
+    VERSION 0.1
+    DEPENDENCIES QtQml
+)
```
outfoxxed: indent is wrong

outfoxxed:
Failing lints
```
/home/admin/programming/outfoxxed/quickshell/i3/src/x11/i3/ipc/qml.cpp:3:1: error: included header qjsonarray.h is not used directly [misc-include-cleaner,-warnings-as-errors]
    3 | #include <qjsonarray.h>
      | ^~~~~~~~~~~~~~~~~~~~~~~
    4 | #include <qobject.h>
```
```
/home/admin/programming/outfoxxed/quickshell/i3/src/x11/i3/ipc/connection.cpp:96:34: error: class member accesses must explicitly use thisptr [tidyfox-explicit-thisptr,-warnings-as-errors]
   96 |         for (auto& [type, data]: I3Ipc::parseResponse()) {
      |                                         ^
      |                                         this->
/home/admin/programming/outfoxxed/quickshell/i3/src/x11/i3/ipc/connection.cpp:135:7: error: no header providing "strncmp" is directly included [misc-include-cleaner,-warnings-as-errors]
    3 |                 if (strncmp(buffer.data(), MAGIC.data(), 6) != 0) {
      |                     ^
/home/admin/programming/outfoxxed/quickshell/i3/src/x11/i3/ipc/connection.cpp:377:13: error: method 'handleRunCommand' can be made static [readability-convert-member-functions-to-static,-warnings-as-errors]
  377 | void I3Ipc::handleRunCommand(I3IpcEvent* event) {
      |             ^
/home/admin/programming/outfoxxed/quickshell/i3/src/x11/i3/ipc/connection.cpp:380:3: error: variable 'success' of type 'bool' can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  380 |                 bool success = obj["success"].toBool();
      |                 ^
      |                      const
/home/admin/programming/outfoxxed/quickshell/i3/src/x11/i3/ipc/connection.cpp:383:4: error: variable 'error' of type 'QString' can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  383 |                         QString error = obj["error"].toString();
      |                         ^
      |                                 const
```

nydragon:
Fixed

outfoxxed (approved):
Thanks!

PR #13, opened 2025-01-18
desktopentry: Fix string parsing in exec string

phonetic112:
makes the inline string detector thing also check single quotes


outfoxxed:
Forgot about this one, mb

PR #22, opened 2025-03-02
cli: add `callJson` ipc subcommand

jmoggr:
The `arguments` parser for the `ipc call` subcommand `allow_extra_args` does not gracefully handle many JSON strings. It will allow '{"hello": "world"}', but it will strip out square brackets and split the JSON string on commas, among other issues.

This is worked around by adding a new `callJson` ipc subcommand which accepts one string argument which is passed directly to the ipc call.

I understand that this is kind of hacky, and am open to modify this or taking a different approach. I do think something should be done for JSON because it can be somewhat frustrating to debug/learn when simple cases work but complex cases fail. At the very least I could ditch this change and update the documentation for `arguments`.

### DIFF
--- a/src/launch/command.cpp
+++ b/src/launch/command.cpp
@@ -295,6 +295,13 @@ int ipcCommand(CommandState& cmd) {
 			return qs::io::ipc::comm::queryMetadata(&client, *cmd.ipc.target, *cmd.ipc.name);
 		} else if (*cmd.ipc.getprop) {
 			return qs::io::ipc::comm::getProperty(&client, *cmd.ipc.target, *cmd.ipc.name);
+		} else if (*cmd.ipc.callJson) {
+			return qs::io::ipc::comm::callFunction(
+			    &client,
+			    *cmd.ipc.target,
+			    *cmd.ipc.name,
+			    {*cmd.ipc.jsonArgument}
+			);
 		} else {
 			QVector<QString> arguments;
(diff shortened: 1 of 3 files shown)


outfoxxed:
This doesn't do anything differently than a normal call, which already don't strip out square brackets or split on commas.

jmoggr:
> which already don't strip out square brackets or split on commas.

That is not the behavior I am experiencing

splitting on commas:
 qs ipc call jsonTarget sendJson '[{"a": 1},{"b":2}]'
Too many arguments provided (1 required but 2 were provided.)
Function definition: function sendJson(jsonString: string): void

stripping brackets:
 qs ipc call jsonTarget sendJson '[]'
Too few arguments provided (1 required but 0 were provided.)
Function definition: function sendJson(jsonString: string): void

code:
```
import Quickshell
import Quickshell.Io

ShellRoot {
    IpcHandler {
        target: "jsonTarget"

        function sendJson(jsonString: string): void {
            console.log(jsonString);
        }
    }
}
```

outfoxxed:
Looked into it. Appears to be a misfeature in the argument parser we use when parsing lists, which can't be disabled. Note that it will only trigger when the first character is '[' and the last character is ']'. I recommend wrapping it in an object for now to avoid those characters.

As it stands I don't want to merge this because it should just work without another subcommand.

PR #23, opened 2025-03-03
musl: fix building with musl linked libc++

burein-ita:
Fixes some musl related quirks.


outfoxxed:
Please leave a comment above the #undefs explaining the issue and when it applies so it doesn't get unintentionally removed later. (force push)

burein-ita:
Donzo

PR #41, opened 2025-05-26
services/pipewire: expose type of node

nydragon:
I wanted to expose the node type, as a first effort to be able to implement a privacy indicator (usage of microphone/camera/screenshare).

### DIFF
--- a/src/services/pipewire/defaults.cpp
+++ b/src/services/pipewire/defaults.cpp
@@ -146,7 +146,7 @@ void PwDefaultTracker::onNodeDestroyed(QObject* node) {
 
 void PwDefaultTracker::changeConfiguredSink(PwNode* node) {
 	if (node != nullptr) {
-		if (!node->isSink) {
+		if (!node->type.testFlags(PwNodeType::AudioSink)) {
 			qCCritical(logDefaults) << "Cannot change default sink to a node that is not a sink.";
 			return;
 		}
@@ -168,7 +168,7 @@ void PwDefaultTracker::changeConfiguredSinkName(const QString& sink) {
 
 void PwDefaultTracker::changeConfiguredSource(PwNode* node) {
 	if (node != nullptr) {
-		if (node->isSink) {
+		if (!node->type.testFlags(PwNodeType::AudioSource)) {
 			qCCritical(logDefaults) << "Cannot change default source to a node that is not a source.";
 			return;
 		}
--- a/src/services/pipewire/node.cpp
+++ b/src/services/pipewire/node.cpp
@@ -11,6 +11,7 @@
 #include <qlogging.h>
 #include <qloggingcategory.h>
 #include <qobject.h>
+#include <qstringliteral.h>
 #include <qtmetamacros.h>
 #include <qtypes.h>
 #include <spa/node/keys.h>
@@ -85,6 +86,20 @@ QString PwAudioChannel::toString(Enum value) {
 	}
 }
 
+QString PwNodeType::toString(PwNodeType::Flags type) {
+	switch (type) {
+	case PwNodeType::VideoSource: return QStringLiteral("VideoSource");
+	case PwNodeType::VideoSink: return QStringLiteral("VideoSink");
+	case PwNodeType::AudioSource: return QStringLiteral("AudioSource");
+	case PwNodeType::AudioSink: return QStringLiteral("AudioSink");
+	case PwNodeType::AudioDuplex: return QStringLiteral("AudioDuplex");
+	case PwNodeType::AudioOutStream: return QStringLiteral("AudioOutStream");
+	case PwNodeType::AudioInStream: return QStringLiteral("AudioInStream");
+	case PwNodeType::Untracked: return QStringLiteral("Untracked");
+	default: return QStringLiteral("Invalid");
+	}
+}
+
 void PwNode::bindHooks() {
 	// Bind the device first as pw is in order, meaning the device should be bound before
 	// we want to do anything with it.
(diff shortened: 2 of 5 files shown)


src/services/pipewire/node.cpp
```diff
@@ -116,21 +129,29 @@ void PwNode::unbindHooks() {
...
 			this->isSink = true;
 			this->isStream = true;
+		} else if (strcmp(mediaClass, "Video/Sink") == 0) {
+			this->type = PwNodeType::VideoSink;
+			this->isSink = false;
+			this->isStream = false;
+		} else if (strcmp(mediaClass, "Video/Source") == 0) {
+			this->type = PwNodeType::VideoSource;
+			this->isSink = false;
+			this->isStream = false;
```
nydragon: I omitted `Audio/Duplex` as I was unsure what combination of isSink and isStream it should be

outfoxxed: Type is meant to indicate the main class of the device, so "Audio" and "Video" here, not including the direction or stream status. Judging by the [pipewire documentation](https://docs.pipewire.org/page_man_pipewire-props_7.html), it doesn't appear pipewire exposes if a video is a stream or not, which is a little unfortunate. I'm not sure if there are any good heuristics to determine it either.

I've never seen an Audio/Duplex, and not sure what it would be so I'm fine leaving it out for now.

nydragon: So should I be renaming this enum to `PwNodeMediaClass`/`PwNodeClass` and add helper functions `isAudio`, `isVideo` or have PwNodeClass and PwNodeType simultaneously ?

I found using the boolean combination isStream and isSink to be a bit counter intuitive, so I'd greatly appreciate something that'd report the pipewire class more closely

outfoxxed: I'd be ok with a bitflag set replacing the booleans as long as we still expose the booleans through qml.

e.g. `Audio | Video | Stream | Source | Sink`.

nydragon: Will update the Pr with that then, but does it make more sense over an enum considering many of the combinations are invalid?

outfoxxed: Flags are useful for testing imo, but you can have both.
e.g. `AudioInStream = Audio | Sink | Stream`. Enums allow multiple definitions for the same value.

nydragon:
~~There is *something* in this PR that makes QS crash on reloads.~~
Nvm rebase fixed it.

src/services/pipewire/defaults.cpp
```diff
@@ -146,7 +146,7 @@ void PwDefaultTracker::onNodeDestroyed(QObject* node) {
 
 void PwDefaultTracker::changeConfiguredSink(PwNode* node) {
 	if (node != nullptr) {
-		if (!node->isSink) {
+		if (node->type != PwNodeType::AudioSink) {
```
outfoxxed: This prevents an application from being set as the default due to apps having the Stream flag. Instead use [`testFlags`](https://doc.qt.io/qt-6/qflags.html#testFlags). `node->type.testFlags(PwNodeType::AudioSink)` should do it.

src/services/pipewire/defaults.cpp
```diff
@@ -168,7 +168,7 @@ void PwDefaultTracker::changeConfiguredSinkName(const QString& sink) {
 
 void PwDefaultTracker::changeConfiguredSource(PwNode* node) {
 	if (node != nullptr) {
-		if (node->isSink) {
+		if (node->type != PwNodeType::AudioSource) {
```
outfoxxed: Same as above.

src/services/pipewire/node.cpp
```diff
@@ -164,7 +175,7 @@ void PwNode::initProps(const spa_dict* props) {
 		}
 	}
 
-	if (this->type == PwNodeType::Audio) {
+	if (this->type & PwNodeType::Audio) {
```
outfoxxed: Use testFlag as its easier to verify at a glance, especially since our flag set contains combined values.

src/services/pipewire/node.hpp
```diff
@@ -86,10 +86,46 @@ class PwAudioChannel: public QObject {
...
+	QML_ELEMENT;
+	QML_SINGLETON;
 
-enum class PwNodeType : quint8 {
-	Untracked,
-	Audio,
+public:
+	enum Enum : quint8 {
+		// A Pipewire node which is not being managed.
+		Untracked = 0x1,
```
outfoxxed: This should be 0x0, as for the others I've switched to using 0b for flags as its easier to verify.
e.g. `0b1`, `0b10`, `0b100`.

src/services/pipewire/qml.cpp
```diff
@@ -328,12 +328,14 @@ QString PwNodeIface::description() const { return this->mNode->description; }
 
 QString PwNodeIface::nickname() const { return this->mNode->nick; }
 
-bool PwNodeIface::isSink() const { return this->mNode->isSink; }
+bool PwNodeIface::isSink() const { return this->mNode->type & PwNodeType::Sink; }
 
-bool PwNodeIface::isStream() const { return this->mNode->isStream; }
+bool PwNodeIface::isStream() const { return this->mNode->type & PwNodeType::Stream; }
```
outfoxxed: testFlag as above

src/services/pipewire/node.hpp
```diff
@@ -86,10 +86,46 @@ class PwAudioChannel: public QObject {
...
+		// A node that is both a sink and a source.
+		AudioDuplex = Audio | Sink | Source,
+		// A playback stream.
+		AudioOutStream = Audio | Sink | Stream,
+		// A capture stream.
+		AudioInStream = Audio | Source | Stream,
+		// A producer of video, like a webcam or a screenshare.
+		VideoSource = Video | Source,
+		// A consumer of video, such as a program that is recieving a video stream.
+		VideoSink = Video | Sink,
```
outfoxxed: Note which pipewire media class each of these is, and what they're equal to, as the definitions don't currently show up in docs.

outfoxxed (changes requested):

src/services/pipewire/defaults.cpp
nydragon: Neat, thanks

src/services/pipewire/node.hpp
nydragon: Added that, let me know if its fine by you

src/services/pipewire/node.hpp
```diff
@@ -86,12 +87,62 @@ class PwAudioChannel: public QObject {
...
+		// and is composed of the @@PwNodeType.Audio, @@PwNodeType.Source and @@PwNodeType.Stream flags.
+		AudioInStream = Audio | Source | Stream,
+		// A producer of video, like a webcam or a screenshare. This is
+		// equivalent to the media class `Video/Source` and is composed of the
+		// @@PwNodeType.Video and @@PwNodeType.Source flags.
+		VideoSource = Video | Source,
+		// A consumer of video, such as a program that is recieving a video stream.
+		// This is equivalent to the media class `Video/Sink` and is composed of the
+		// @@PwNodeType.Video and @@PwNodeType.Sink flags.
+		VideoSink = Video | Sink,
```
outfoxxed: Put an empty /// line between the equivalent to and description lines.

src/services/pipewire/node.hpp
```diff
@@ -86,12 +87,62 @@ class PwAudioChannel: public QObject {
 	/// including aux and custom channel ranges.
 	Q_INVOKABLE static QString toString(qs::service::pipewire::PwAudioChannel::Enum value);
 };
+///! The type of a pipewire node. Use bitwise comparisons to
```
outfoxxed: newline after ;

outfoxxed (changes requested):
Looks good apart from these nits. Will test soon.

src/services/pipewire/node.hpp
nydragon: Fixed :+1:

Issue #74, opened 2025-06-24
NetworkManager service


cpwrs:
Is there any plans/tracking for a NetworkManager service? I just started using quickshell and noticed many configs defer to an nmcli Process which is a pain to parse. 

I'm interested in implementing the service using the [D-Bus API](https://networkmanager.dev/docs/api/latest/spec.html). The spec is large so I'd start with a subset relevant to Wireless.

outfoxxed:
There are plans yes. @fufexan was working on it at one point, not sure if he still is.

Feel free to take a crack at it.

fufexan:
[nm.patch.txt](https://github.com/user-attachments/files/20898092/nm.patch.txt)

Here's the single commit I have. Not sure it even works, but you may find some value in the scaffolding.

cpwrs:
Thanks! I'll check that out

vesim987:
Hi, there is [networkmanager-qt](https://github.com/KDE/networkmanager-qt) already, any thought on using it? The needed dependencies feel reasonable.

outfoxxed:
There is ongoing work already. New dependencies are not permitted in most cases.

PR #96, opened 2025-07-06
NetworkManager Service

cpwrs:
Work in progress NetworkManager service.
Starting with a subset of the API relevant to exposing Device, Wireless, AccessPoint, and Connection objects via DBus. Spec: https://networkmanager.dev/docs/api/latest/spec.html

To do:

org.freedesktop.NetworkManager
- [X] NetworkManager class
- [X] Connect to DBus
- [X] Connect to NM service
- [X] Track devices (GetAllDevices, DeviceAdded, DeviceRemoved)
- [X] Track a default Wifi device
- [X] Track state property

org.freedesktop.NetworkManager.Device
- [X] NMDevice class
- [X] Track Path, Interface, DeviceType, Managed properties
- [ ] Track ActiveConnection object

org.freedesktop.NetworkManager.Device.Wireless
 - [X] Find wifi devices
 - [ ] NMWirelessDevice class
 - [ ] Track access point objects (AccessPoints, ActiveAccessPoint)
 - [ ] Add trigger for access point rescans
 - [ ] AddAndActivateConnection method (Needs consideration of secret agent for password entry)
<br>

- [ ] Polish Qml API
- [ ] Document and test


outfoxxed:
Why are we linking nm? Should be entirely over dbus

outfoxxed:
I'm not sure what stage of development you're at here and haven't checked code yet, but just so we're clear I'd like to expose this under `Quickshell.Network` and keep the API clean of NM specific things (an IWD backend has been requested as well).

I don't think I've made that clear so far. (If you've got any questions, @ me here or in the matrix/discord)

domenkozar:
Really looking forward to this one! :)

cpwrs:
Still missing a secret agent for networks that require passwords. I've tested with nm-applet as the agent and everything seems to work. 

I completely separated the frontend and backend objects, with some nm-specific members in the frontend. Happy to do multi-backend in a different style though.

I think this is enough features (+ agent) for most WiFi use cases, but there's a ton more functionality to add in the future:
- Connection manager
- VPN connections
- Wired/ethernet
- Device speed

src/network/network.hpp
```diff
@@ -0,0 +1,68 @@
...
+/// which can be used to view, configure, and connect to various networks.
+class Network: public QObject {
+	Q_OBJECT;
+	QML_NAMED_ELEMENT(Network);
+	QML_SINGLETON;
+
+	/// The wifi device service.
+	Q_PROPERTY(Wifi* wifi READ wifi CONSTANT);
+	/// The backend being used to power the Network service.
+	Q_PROPERTY(NetworkBackendType::Enum backend READ backend);
```
outfoxxed: The namespace needs to be included for types in a Q_PROPERTY definition.

src/network/wifi.hpp
```diff
@@ -0,0 +1,160 @@
...
+	void nmReasonChanged();
+	void nmSecurityChanged();
+	void requestConnect();
+
+public slots:
+	void setSignalStrength(quint8 signalStrength);
+	void setState(NetworkConnectionState::Enum state);
+	void setNmReason(NMConnectionStateReason::Enum reason);
+	void setNmSecurity(NMWirelessSecurityType::Enum security);
+	void setKnown(bool known);
```
outfoxxed: Why are these slots public? These can be invoked from QML.

src/network/nm/backend.cpp
```diff
@@ -0,0 +1,224 @@
...
+	qDBusRegisterMetaType<ConnectionSettingsMap>();
+
+	auto bus = QDBusConnection::systemBus();
+	if (!bus.isConnected()) {
+		qCWarning(logNetworkManager
+		) << "Could not connect to DBus. NetworkManager backend will not work.";
+		return;
+	}
+
+	this->proxy = new DBusNetworkManagerProxy(NM_SERVICE, NM_PATH, bus, this);
```
outfoxxed: inline service/path

src/network/nm/backend.cpp
```diff
@@ -0,0 +1,224 @@
...
+	QObject::connect(wireless, &NMWirelessDevice::stateChanged, device, [device](NMDeviceState::Enum state) { device->setState(qs::network::NetworkManager::toNetworkDeviceState(state));});
+	QObject::connect(wireless, &NMWirelessDevice::lastScanChanged, device, &WifiDevice::scanComplete);
+	QObject::connect(wireless, &NMWirelessDevice::addAndActivateConnection, this, &NetworkManager::addAndActivateConnection);
+	QObject::connect(wireless, &NMWirelessDevice::activateConnection, this, &NetworkManager::activateConnection);
+	QObject::connect(wireless, &NMWirelessDevice::wifiNetworkAdded, device, &WifiDevice::networkAdded);
+	QObject::connect(wireless, &NMWirelessDevice::wifiNetworkRemoved, device, &WifiDevice::networkRemoved);
+	QObject::connect(device, &WifiDevice::requestScan, wireless, &NMWirelessDevice::scan);
+	// clang-format on
+
+	emit wifiDeviceAdded(device);
```
outfoxxed: this->

src/network/nm/accesspoint.cpp
```diff
@@ -0,0 +1,71 @@
...
+DBusDataTransform<qs::network::NM80211ApSecurityFlags::Enum>::fromWire(quint32 wire) {
+	return DBusResult(static_cast<qs::network::NM80211ApSecurityFlags::Enum>(wire));
+}
+
+DBusResult<qs::network::NM80211Mode::Enum>
+DBusDataTransform<qs::network::NM80211Mode::Enum>::fromWire(quint32 wire) {
+	return DBusResult(static_cast<qs::network::NM80211Mode::Enum>(wire));
+}
+
+} // namespace qs::dbus
```
outfoxxed: Leave the dbus definitions at the bottom please

src/network/nm/accesspoint.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+	if (!this->proxy->isValid()) {
+		qCWarning(logNetworkManager) << "Cannot create DBus interface for AP at" << path;
+		return;
+	}
+
+	// clang-format off
+	QObject::connect(&this->accessPointProperties, &DBusPropertyGroup::getAllFinished, this, [this]() { emit this->ready(); }, Qt::SingleShotConnection);
+	bSecurity.setBinding([&] { return findBestWirelessSecurity( this->mCaps, true, this->bMode == NM80211Mode::Adhoc, this->bFlags, this->bWpaFlags, this->bRsnFlags); });
+	// clang-format on
```
outfoxxed: don't clang-format this one

src/network/nm/connection.cpp
```diff
@@ -0,0 +1,119 @@
...
+	if (!this->proxy->isValid()) {
+		qCWarning(logNetworkManager) << "Cannot create DBus interface for connection at" << path;
+		return;
+	}
+
+	// clang-format off
+	QObject::connect(this->proxy, &DBusNMConnectionSettingsProxy::Updated, this, &NMConnectionSettings::updateSettings);
+	bSecurity.setBinding([&] { return securityFromConnectionSettings(this->bSettings); });
+	// clang-format on
+	//
```
outfoxxed: //?

src/network/nm/connection.cpp
```diff
@@ -0,0 +1,119 @@
...
+	    this
+	);
+
+	if (!this->proxy->isValid()) {
+		qCWarning(logNetworkManager) << "Cannot create DBus interface for connection at" << path;
+		return;
+	}
+
+	// clang-format off
+	QObject::connect(&this->activeConnectionProperties, &DBusPropertyGroup::getAllFinished, this, [this]() { emit this->ready(); }, Qt::SingleShotConnection);
```
outfoxxed: You can bind this directly to ::ready instead of a lambda.

src/network/nm/device.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+	if (!this->deviceProxy->isValid()) {
+		qCWarning(logNetworkManager) << "Cannot create DBus interface for device at" << path;
+		return;
+	}
+
+	// clang-format off
+	QObject::connect(this, &NMDevice::availableConnectionPathsChanged, this, &NMDevice::onAvailableConnectionPathsChanged);
+	QObject::connect(this, &NMDevice::activeConnectionPathChanged, this, &NMDevice::onActiveConnectionPathChanged);
+	QObject::connect(&this->deviceProperties, &DBusPropertyGroup::getAllFinished, this, [this]() { emit this->deviceReady(); }, Qt::SingleShotConnection);
```
outfoxxed: direct bind

src/network/nm/utils.cpp
```diff
@@ -0,0 +1,341 @@
...
+		}
+		if (haveAp && !(apRsn & NM80211ApSecurityFlags::KeyMgmtEapSuiteB192)) {
+			return false;
+		}
+		break;
+	default: good = false; break;
+	}
+
+	return good;
+}
```
outfoxxed, line 207: I'm assuming this is copied directly from the NM source, but I think it would be beneficial to rewrite it in a way that is intended to be read. At least removing braces for single-line ifs would go a long way.

src/network/nm/wireless.cpp
```diff
@@ -0,0 +1,383 @@
...
+	if (this->mReason == reason) return;
+	this->mReason = reason;
+	emit this->reasonChanged(reason);
+};
+
+void NMWirelessNetwork::setKnown(bool known) {
+	if (this->mKnown == known) return;
+	this->mKnown = known;
+	emit this->knownChanged(known);
+}
```
outfoxxed: At a glance things like this may be better handled by bindables, in general

src/network/nm/wireless.cpp
```diff
@@ -0,0 +1,383 @@
...
+		}
+	}
+};
+
+void NMWirelessNetwork::addConnectionSettings(NMConnectionSettings* conn) {
+	if (this->mConnections.contains(conn->path())) return;
+	this->mConnections.insert(conn->path(), conn);
+	// clang-format off
+	QObject::connect(conn, &NMConnectionSettings::securityChanged, this, &NMWirelessNetwork::updateReferenceConnection);
+	QObject::connect(conn, &NMConnectionSettings::disappeared, this, [this, conn]() { this->removeConnectionSettings(conn); });
```
outfoxxed: In general I prefer to cast the signal sender instead of passing a lambda to connect for cases like this.

src/network/nm/wireless.cpp
```diff
@@ -0,0 +1,383 @@
...
+	// clang-format on
+	this->updateReferenceConnection();
+	this->setKnown(true);
+};
+
+void NMWirelessNetwork::removeConnectionSettings(NMConnectionSettings* conn) {
+	auto* found = this->mConnections.take(conn->path());
+	if (found) {
```
outfoxxed: Inline found into the if, same for similar cases

src/network/device.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+QString NetworkConnectionState::toString(NetworkConnectionState::Enum state) {
+	switch (state) {
+	case Unknown: return "Unknown";
+	case Connecting: return "Connecting";
+	case Connected: return "Connected";
+	case Disconnecting: return "Disconnecting";
+	case Disconnected: return "Disconnected";
+	}
+	return {};
```
outfoxxed: Prefer "Invalid"

src/network/test/network.qml
```diff
@@ -0,0 +1,169 @@
+import QtQuick
```
outfoxxed: I'd prefer if you made the wifi tester a little more usable like the bluetooth one, so its easier to regression test with.

outfoxxed (changes requested):
Sorry its taken so long but here's an initial review.

I'd appreciate if auto was used in more places, as well as whitespace following the existing code-style more as well.

There are a couple bits in here that look completely insane, but judging by the fact that they're on the NM api surface I'm going to assume it's NM's fault.

src/network/nm/utils.cpp
cpwrs: Don't think we will have a use case for haveAp so I removed it. Fixed style too.

cpwrs:
@outfoxxed Thanks for the feedback! I think this is ready for another review.

> There are a couple bits in here that look completely insane, but judging by the fact that they're on the NM api surface I'm going to assume it's NM's fault.

Yep. Unfortunately the api doesn't have any concept of wifi networks. Had to grab access points from one interface, "active" and "saved" connections from another, and then group them by SSID. It also doesn't give any useful security info, hence the copied code from nmcli.

Other clients do something similar:
https://invent.kde.org/plasma/plasma-nm/-/blob/master/libs/models/networkmodel.cpp?ref_type=heads
https://invent.kde.org/frameworks/networkmanager-qt/-/blob/master/src/wirelessnetwork.cpp?ref_type=heads

cpwrs:
@bbedward Would love your feedback on this API. Besides the obvious missing features, does the design make anything difficult for DMS?

bbedward:
> @bbedward Would love your feedback on this API. Besides the obvious missing features, does the design make anything difficult for DMS?

It looks pretty consistent with quickshell conventions to me. I'd like to see the agent/connection manager, but it's not necessary for  the initial integration IMO (the bluez implementation doesn't have an agent either)

src/network/network.hpp
```diff
@@ -0,0 +1,62 @@
...
+///! The Network service.
+/// An interface to a network backend (currently only NetworkManager),
+/// which can be used to view, configure, and connect to various networks.
+class Network: public QObject {
+	Q_OBJECT;
+	QML_SINGLETON;
+	QML_NAMED_ELEMENT(Network);
+
+	/// The wifi device service.
+	Q_PROPERTY(qs::network::Wifi* wifi READ wifi CONSTANT);
```
outfoxxed: Why are all networks exposed under wifi? What about things like ethernet, wireguard tunnels, etc? Its fine not to have those in the initial PR but the API should be reasonably extensible to them.

Note that its also entirely possible to have more than one wireless interface at once, though I haven't checked how NM exposes that.

src/network/wifi.hpp
```diff
@@ -0,0 +1,151 @@
...
+	/// True if the network is connected.
+	Q_PROPERTY(bool connected READ connected NOTIFY connectedChanged BINDABLE bindableConnected);
+	/// True if the wifi network has known connection settings saved.
+	Q_PROPERTY(bool known READ known NOTIFY knownChanged BINDABLE bindableKnown);
+	/// A specific reason for the connection state when the backend is NetworkManager.
+	Q_PROPERTY(NMConnectionStateReason::Enum nmReason READ nmReason NOTIFY nmReasonChanged BINDABLE bindableNmReason);
+	/// The security type of the wifi network when the backend is NetworkManager.
+	Q_PROPERTY(NMWirelessSecurityType::Enum nmSecurity READ nmSecurity NOTIFY nmSecurityChanged BINDABLE bindableNmSecurity);
+	// clang-format on
+
```
outfoxxed, line 107: I think a class for specifically a wifi network makes sense, but there should be a more general network base class as described above, with properties like a given name and connection status.

src/network/wifi.hpp
```diff
@@ -0,0 +1,151 @@
...
+public slots:
+	void networkAdded(WifiNetwork* net);
+	void networkRemoved(WifiNetwork* net);
+
+public:
+	explicit WifiDevice(QObject* parent = nullptr);
+
+	/// Request the wireless device to scan for available WiFi networks.
+	/// This should be invoked everytime you want to show the user an accurate list of available networks.
+	Q_INVOKABLE void scan();
```
outfoxxed: Regardless of how NM in particular works I question if a scan request is a good interface. My preferred design is to have another session object of sorts keep a scan alive, which keeps it both reactive and hard to forget about as an api consumer.

src/network/wifi.hpp
```diff
@@ -0,0 +1,151 @@
...
+	Q_PROPERTY(quint8 signalStrength READ signalStrength NOTIFY signalStrengthChanged BINDABLE bindableSignalStrength);
+	/// True if the network is connected.
+	Q_PROPERTY(bool connected READ connected NOTIFY connectedChanged BINDABLE bindableConnected);
+	/// True if the wifi network has known connection settings saved.
+	Q_PROPERTY(bool known READ known NOTIFY knownChanged BINDABLE bindableKnown);
+	/// A specific reason for the connection state when the backend is NetworkManager.
+	Q_PROPERTY(NMConnectionStateReason::Enum nmReason READ nmReason NOTIFY nmReasonChanged BINDABLE bindableNmReason);
+	/// The security type of the wifi network when the backend is NetworkManager.
+	Q_PROPERTY(NMWirelessSecurityType::Enum nmSecurity READ nmSecurity NOTIFY nmSecurityChanged BINDABLE bindableNmSecurity);
+	// clang-format on
```
outfoxxed: Do non-nm wireless APIs not expose network security mode? If so which?

src/network/wifi.hpp
```diff
@@ -0,0 +1,151 @@
...
+///! An available wifi network.
+class WifiNetwork: public QObject {
+	Q_OBJECT;
+	QML_ELEMENT;
+	QML_UNCREATABLE("WifiNetwork can only be acquired through WifiDevice");
+	// clang-format off
+	/// The SSID (service set identifier) of the wifi network.
+	Q_PROPERTY(QString ssid READ ssid CONSTANT);
+	/// The current signal strength of the network, in percent.
+	Q_PROPERTY(quint8 signalStrength READ signalStrength NOTIFY signalStrengthChanged BINDABLE bindableSignalStrength);
```
outfoxxed: In percent or ranged 0.0 to 1.0? Conventionally most things in qs are 0.0 to 1.0 with unspecified precision, so it can be displayed with whatever accuracy is desired.

src/network/test/manual/network.qml
```diff
@@ -0,0 +1,132 @@
...
+						}
+					}
+
+					Repeater {
+						Layout.fillWidth: true
+						model: sortedNetworks
+
+						WrapperRectangle {
+							Layout.fillWidth: true
+							color: modelData.connected ? "lightsteelblue" : palette.button
```
outfoxxed: Would change to highlight color or something

src/network/test/manual/network.qml
```diff
@@ -0,0 +1,132 @@
...
+				}
+				CheckBox {
+					enabled: false
+					text: "Hardware"
+					checked: Network.wifi.hardwareEnabled
+				}
+			}
+		}
+
+		ListView {
```
outfoxxed, line 36: clip this

src/network/network.hpp
cpwrs: Wifi is just a container for an `ObjectModel<WifiDevice>*` (the multiple interfaces you refer to) and other wifi properties. It doesn't have state or represent anything on the backend.

I'll move its contents to the top level:
Network.wifiDevices
Network.wifiEnabled
Network.wifiHardwareEnabled

src/network/wifi.hpp
cpwrs: They all expose some security. IWD for example is one of "open", "wep", "psk", or "8021x". 

NMWirelessSecurityType is relevant to NM users (especially once a connection editor is added), and keeping them seperate would avoid a lossy mapping.

src/network/wifi.hpp
cpwrs: NM provides an int8 percentage, IWD provides int16 100 * dBm.
I can't get reverse percent -> dBm without losing precision.

I'll make it 0.0-1.0 for all and just keep the extra precision other backends give.

src/network/wifi.hpp
outfoxxed: Is there a reason to avoid making the exposed connection type list a superset of what all backends provide?

src/network/network.hpp
outfoxxed: I think this is too general for some cases. Network is in a similar position to bluetooth where you can have multiple adapters with different devices, settings, and active connection limits. It is entirely possible to have one interface connected to 3 networks while another is rfkilled, afaik.

The design I took with the bluetooth module was to represent the physical adapter->device layout but also provide a toplevel list that spans all adapters. I think it would make sense for NM to take a similar direction here API wise. If you're opposed to an approach like that please elaborate on it.

cpwrs: NM abstracts away the adapter(phy) and lacks per-adapter/per-device radio controls. Sadly it only supports global rfkill state for Wifi and WWAN (flip all software or hardware switches). 

IWD has a great API and exposes all this stuff so I agree with the approach.

I'd need to use sysfs to organize interfaces by their adapter for NM backend.
`/sys/class/net/${iface}/phy80211/name` gives `phy{#}`.
And then NM backend wouldn't implement any of the adapter methods.
Too hacky?

src/network/wifi.hpp
cpwrs: That should work!

src/network/network.hpp
outfoxxed: Thats quite unfortunate. With that restriction just exposing the network list and toplevel controls is a reasonable starting point, though I don't like the wifi categorization.

src/network/wifi.hpp
cpwrs: Could you elaborate on what this would look like?

NetworkManager scan is a shot in the dark. It can silently fail, rate limit, etc. The success indicator is a last scan timestamp right now. The NM access points list removes old or invalid APs, scan is just to search for new ones.

cpwrs:
Resolved all but scan

src/network/wifi.hpp
outfoxxed: Bluez is somewhat similar, though I didnt figure out the pattern I wanted before writing that, so its not there.

The idea is to make a scanner object which has an enable property. When enabled, the scanner performs all the underlying work to actively scan, in this case likely just running scan() at some reasonable interval (if we can't figure out a reasonable interval how is a random user supposed to).

The reason I like this pattern is it ties the scan being active to the lifecycle of the object which makes it hard to forget about (as opposed to the current bluetooth implementation which is very easy to forget).

I am undecided on if newly discovered entries should be exposed exclusively through the scanner object while enabled or available in the normal list, but I don't like them being exposed while not scanning at least. I would hide those if there is a clear indicator they are only being exposed due to a random scan.

cpwrs: That sounds good. Maybe I'd remove WifiNetwork.connected and move to WifiDevice.connectedNetwork and WifiDevice.availableNetworks (empty unless Scanning.enabled).

outfoxxed: I think leaving .connected available is a good idea for two reasons
1. Remembered networks that arent currently connected
2. I expect a lot of people would like to use the same delegate for both a connected and disconnected network

As for splitting apart the lists, we could do it in any way without much issue because users can use ScriptModel to filter or combine models, but keeping it all in one list is not unreasonable.

outfoxxed: As a side note for the NM scan call, it might be beneficial to check how and when KDE calls it, since they're probably doing it correctly and do repeated scans when the network panel is open IIRC.

cpwrs:
Made an attempt at WifiScanner following the KDE implementation. Requires deprecated header `<time.h>` for `CLOCK_BOOTTIME` -> `QDateTime`. I like the scanner but the conditional network visibility is inherently hacky because NM. Kept it in one list, networks are visible if scanning or connected or known.

outfoxxed:
Sorry I haven't looked at this recently, is it ready for review again?

cpwrs:
@outfoxxed Yep! 

I decided to keep WifiNetwork as a group of all access points and connection settings for a given SSID. KDE shows a network item PER conn settings, which makes sense because its only backend (NM) is connection centric. But I think 99% of use is connecting to a network which only has one conn settings. IWD doesn't even allow multiple. Right now, if theres multiple it attempts a connection using the best settings (highest security). 

Eventually we'd want to expose the AccessPoint(s) and ConnSetting(s) (only one for IWD) under these networks and expand the connect() function to optionally take a specific AP or settings. This gives feature parity with the NM API for the 1% who actually use multiple settings. 

A connection editor and these extensions to WifiNetwork would be a large feature so I was hoping to get the basics merged before that. The biggest issue with not including ConnSetting initially is losing a forget() function that removes the setting. Does a forget() at the WifiNetwork level that deletes ALL its associated settings suffice for now?

src/network/wifi.hpp
```diff
@@ -0,0 +1,203 @@
...
+	QBindable<bool> bindableEnabled() { return &this->bEnabled; };
+	[[nodiscard]] bool enabled() const { return this->bEnabled; };
+	void setEnabled(bool enabled);
+
+signals:
+	void enabledChanged(bool enabled);
+
+private:
+	Q_OBJECT_BINDABLE_PROPERTY(WifiScanner, bool, bEnabled, &WifiScanner::enabledChanged);
+};
```
cpwrs: This can reduce to a bool property unless there's more functionality to add?

outfoxxed: At this point yeah

src/network/network.hpp
```diff
@@ -0,0 +1,113 @@
...
+	// clang-format off
+	/// A list of all network devices.
+	Q_PROPERTY(UntypedObjectModel* devices READ devices CONSTANT);
+	QSDOC_TYPE_OVERRIDE(ObjectModel<qs::Network::Device>*);
+	/// The backend being used to power the Network service.
+	Q_PROPERTY(qs::network::NetworkBackendType::Enum backend READ backend CONSTANT);
+	/// Switch for the rfkill software block of all wireless devices.
+	Q_PROPERTY(bool wifiEnabled READ wifiEnabled WRITE setWifiEnabled NOTIFY wifiEnabledChanged);
+	/// Switch for the rfkill hardware block of all wireless devices.
+	Q_PROPERTY(bool wifiHardwareEnabled READ wifiHardwareEnabled NOTIFY wifiHardwareEnabledChanged BINDABLE bindableWifiHardwareEnabled);
```
outfoxxed: Wouldn't call this a switch if its readonly

src/network/network.cpp
```diff
@@ -0,0 +1,53 @@
...
+	qCCritical(logNetwork) << "Network will not work. Could not find an available backend.";
+}
+
+void Network::onDeviceAdded(NetworkDevice* dev) { this->mDevices.insertObject(dev); }
+void Network::onDeviceRemoved(NetworkDevice* dev) { this->mDevices.removeObject(dev); }
+
+void Network::setWifiEnabled(bool enabled) {
+	if (this->bWifiEnabled == enabled) {
+		const QString state = enabled ? "enabled" : "disabled";
+		qCCritical(logNetwork) << "Wifi is already globally software" << state;
```
outfoxxed: Duplicate sets are somewhat expected to happen, just ignore them

src/network/device.hpp
```diff
@@ -0,0 +1,122 @@
...
+	QML_UNCREATABLE("Devices can only be acquired through Network");
+	// clang-format off
+	/// The device type.
+	Q_PROPERTY(DeviceType::Enum type READ type CONSTANT);
+	/// The name of the device's control interface.
+	Q_PROPERTY(QString name READ name NOTIFY nameChanged BINDABLE bindableName);
+	/// The hardware address of the device in the XX:XX:XX:XX:XX:XX format.
+	Q_PROPERTY(QString address READ address NOTIFY addressChanged BINDABLE bindableAddress);
+	/// Connection state of the device.
+	Q_PROPERTY(qs::network::DeviceConnectionState::Enum state READ state NOTIFY stateChanged BINDABLE bindableState);
```
outfoxxed: A "connected" property that just binds to the state and does an eq would be useful

src/network/device.cpp
```diff
@@ -0,0 +1,71 @@
...
+		return QStringLiteral("Waiting for a required secondary connection to activate");
+	case Activated: return QStringLiteral("Connected");
+	case Deactivating: return QStringLiteral("Disconnecting");
+	case Failed: return QStringLiteral("Failed to connect");
+	};
+}
+
+NetworkDevice::NetworkDevice(DeviceType::Enum type, QObject* parent)
+    : QObject(parent)
+    , mType(type) {};
```
outfoxxed: Can go in the header

src/network/wifi.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+///! Wireless variant of a NetworkDevice.
+class WifiDevice: public NetworkDevice {
+	Q_OBJECT;
+	QML_ELEMENT;
+	QML_UNCREATABLE("WifiDevices can only be acquired through Wifi");
+
+	/// A list of this available and connected wifi networks.
+	Q_PROPERTY(UntypedObjectModel* networks READ networks CONSTANT);
+	QSDOC_TYPE_OVERRIDE(ObjectModel<WifiNetwork>*)
```
outfoxxed: We are completely ignoring wired networks which nm supports by putting this in the device layer, but I think we've already talked about this, and its acceptable even without them for now. Just pointing out that I can enable/disable a wired connection, though I'm not sure if you can do anything more fine grained than that.

src/network/wifi.hpp
```diff
@@ -0,0 +1,203 @@
...
+public:
+	explicit WifiNetwork(QString ssid, QObject* parent = nullptr);
+
+	/// Attempt to connect to the wifi network.
+	Q_INVOKABLE void connect();
+
+	QBindable<qreal> bindableSignalStrength() { return &this->bSignalStrength; }
+	[[nodiscard]] qreal signalStrength() const { return this->bSignalStrength; };
+	QBindable<bool> bindableKnown() { return &this->bKnown; }
+	[[nodiscard]] bool known() const { return this->bKnown; };
```
outfoxxed: We should have a way to forget a known network if possible, assuming "known" means "saved"

Also in general if you expose the BINDABLE in the Q_PROPERTY you should make the READ `default` which generates the trivial getter internally.

src/network/wifi.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+	/// A list of this available and connected wifi networks.
+	Q_PROPERTY(UntypedObjectModel* networks READ networks CONSTANT);
+	QSDOC_TYPE_OVERRIDE(ObjectModel<WifiNetwork>*)
+	/// The wifi scanner for this device.
+	Q_PROPERTY(WifiScanner* wifiScanner READ wifiScanner CONSTANT);
+	/// The 802.11 mode the device is in.
+	Q_PROPERTY(WifiDeviceMode::Enum mode READ mode NOTIFY modeChanged BINDABLE bindableMode);
+
+public slots:
```
outfoxxed: Should these slots be public? This makes them callable from QML.

src/network/wifi.cpp
```diff
@@ -0,0 +1,130 @@
...
+	case WpaEap: return QStringLiteral("WPA Enterprise");
+	case WpaPsk: return QStringLiteral("WPA");
+	case StaticWep: return QStringLiteral("WEP");
+	case DynamicWep: return QStringLiteral("Dynamic WEP");
+	case Leap: return QStringLiteral("LEAP");
+	case Owe: return QStringLiteral("OWE");
+	case Open: return QStringLiteral("Open");
+	case Unknown: return QStringLiteral("Unknown");
+	}
+}
```
outfoxxed, line 37: All toString switches should have a default case because random integers can be passed in from QML.

src/network/wifi.cpp
```diff
@@ -0,0 +1,130 @@
...
+	case DeviceRemoved: return QStringLiteral("The device this connection depended on disappeared.");
+	};
+};
+
+WifiScanner::WifiScanner(QObject* parent): QObject(parent) {};
+
+void WifiScanner::setEnabled(bool enabled) {
+	if (this->bEnabled == enabled) {
+		const QString state = enabled ? "enabled" : "disabled";
+		qCCritical(logWifi) << "Scanner is already" << state;
```
outfoxxed: Double sets somewhat expected again

src/network/CMakeLists.txt
outfoxxed, line 1: Any reason not to put the nm stuff in a CMakeLists in its own dir?

src/network/nm/device.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+private slots:
+	void onAvailableConnectionPathsChanged(const QList<QDBusObjectPath>& paths);
+	void onActiveConnectionPathChanged(const QDBusObjectPath& path);
+
+private:
+	void registerConnection(const QString& path);
+
+	QSet<QString> mConnectionPaths;
+	QHash<QString, NMConnectionSettings*> mConnections;
```
outfoxxed: Why are these not one QHash?

src/network/nm/backend.cpp
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	// Create a temporary interface to decide the device type which won't change.
+	auto* temp = new QDBusInterface(
+	    "org.freedesktop.NetworkManager",
+	    path,
+	    "org.freedesktop.DBus.Properties",
+	    QDBusConnection::systemBus(),
+	    this
+	);
```
outfoxxed: asyncReadProperty from dbus/properties.hpp may be nicer

src/network/nm/backend.cpp
```diff
@@ -0,0 +1,265 @@
...
+		const QDBusPendingReply<QVariant> reply = *call;
+		if (reply.isError()) {
+			qCWarning(logNetworkManager) << "Failed to get device type: " << reply.error().message();
+		} else {
+			auto type = static_cast<qs::network::NMDeviceType::Enum>(reply.value().toInt());
+			NMDevice* dev = nullptr;
+
+			switch (type) {
+			case NMDeviceType::Wifi: dev = new NMWirelessDevice(path); break;
+			default: return;
```
outfoxxed: return without deleting call, same with following returns. Probably wrap this / use asyncReadProperty as mentioned above which deletes the callback

src/network/nm/wireless.cpp
```diff
@@ -0,0 +1,442 @@
...
+	auto onDisappeared = [this, ap]() {
+		if (this->mAccessPoints.take(ap->path())) {
+			QObject::disconnect(ap, nullptr, this, nullptr);
+			this->updateReferenceAp();
+			if (this->mAccessPoints.isEmpty() && this->mConnections.isEmpty()) emit this->disappeared();
+		}
+	};
+	// clang-format off
+	QObject::connect(ap, &NMAccessPoint::signalStrengthChanged, this, &NMWirelessNetwork::updateReferenceAp);
+	QObject::connect(ap, &NMAccessPoint::disappeared, this, onDisappeared);
```
outfoxxed: Can an AP be destroyed without disappeared?

src/network/nm/wireless.cpp
```diff
@@ -0,0 +1,442 @@
...
+		if (this->mConnections.take(conn->path())) {
+			QObject::disconnect(conn, nullptr, this, nullptr);
+			this->updateReferenceConnection();
+			if (this->mConnections.isEmpty()) this->bKnown = false;
+			if (this->mAccessPoints.isEmpty() && this->mConnections.isEmpty()) emit this->disappeared();
+		}
+	};
+	// clang-format off
+	QObject::connect(conn, &NMConnectionSettings::securityChanged, this, &NMWirelessNetwork::updateReferenceConnection);
+	QObject::connect(conn, &NMConnectionSettings::disappeared, this, onDisappeared);
```
outfoxxed: Same as above

src/network/nm/wireless.cpp
```diff
@@ -0,0 +1,442 @@
...
+	auto onDisappeared = [this, active]() {
+		if (this->mActiveConnection && this->mActiveConnection == active) {
+			QObject::disconnect(active, nullptr, this, nullptr);
+			this->mActiveConnection = nullptr;
+			this->updateReferenceConnection();
+			this->bState = NMConnectionState::Deactivated;
+			this->bReason = NMConnectionStateReason::None;
+		}
+	};
+	QObject::connect(active, &NMActiveConnection::disappeared, this, onDisappeared);
```
outfoxxed: etc

src/network/nm/backend.cpp
cpwrs: Moved to asyncReadProperty but still requires a temp->deleteLater() I think. I got rid of early returns in the callback. Do you prefer this style?

src/network/wifi.hpp
cpwrs: Sounds good. I planned on eventually exposing "conn settings" in the API and WifiNetwork is just a convenient grouping of all AccessPoints and ConnSettings for a specific network (same ssid).

I don't know if this is convenient for wired and others because it'll just be connect/disconnect (no APs, security settings). Could directly show a list of ConnSettings instead.

This is similar to the difference between 
`nmcli dev wifi connect NETWORK_NAME` and `nmcli conn CONN_SETTINGS_NAME up/down`

Here's what a ConnSettings would look like:
https://git.kernel.org/pub/scm/network/wireless/iwd.git/tree/doc/knownnetwork-api.txt
https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.Settings.Connection.html

Thaodan:
Would it make sense to reuse the KF6 NetworkManagerQt bindings? 

https://api.kde.org/networkmanager-qt-index.html

It's a tier 1 API so the dependencies are very minor, same for eventual ModemManager. Take a look at Plasma-NM for inspiration.

cpwrs:
> Would it make sense to reuse the KF6 NetworkManagerQt bindings?

It's not permitted here, see #74. I don't think there's a good argument to use it over the qt_add_dbus_interface pattern that the other services use.

Thaodan:
Carson Powers ***@***.***> writes:

>> Would it make sense to reuse the KF6 NetworkManagerQt bindings?
>
> It's not permitted here, see #74. I don't think there's a good
> argument to use it over the qt_add_dbus_interface pattern that the
> other services use.

OK that's disappointing. I don't think it's good to reinvent the wheel
here.

In any case I think there is a reason that most projects end of wrapping
libnm, i.e. to reduce the overhead on maintaining their NetworkManager
client API.

The NetworkManager documentation recommends to use LibNM for more
complex cases such as this one:
https://www.networkmanager.dev/docs/libnm/latest/usage.html

Thaodan:
On another does his have to be added as internal module? Maybe it's better if this lives as an external/separately packaged plugin.

outfoxxed:
@Thaodan
> OK that's disappointing. I don't think it's good to reinvent the wheel
here.
>
> In any case I think there is a reason that most projects end of wrapping
libnm, i.e. to reduce the overhead on maintaining their NetworkManager
client API.
>
> The NetworkManager documentation recommends to use LibNM for more
complex cases such as this one:
https://www.networkmanager.dev/docs/libnm/latest/usage.html

> On another does his have to be added as internal module? Maybe it's better if this lives as an external/separately packaged plugin.

Eventually a lot of the project should probably be split into separately installable modules, however that requires some work I haven't had any time to do. If it *were* more modular I wouldn't have a problem with an nm dep for nm, though my general policy against KDE deps remains due to kde/distro release cycles, some distros effectively forcing that pulling in one kde dep pulls in all of kde.

@cpwrs I will get to this this weekend. I've had a lot going on and haven't had time but I want to apologize because the review cycle for this has been far too long.

Thaodan:
outfoxxed ***@***.***> writes:

> outfoxxed left a comment (quickshell-mirror/quickshell#96)
>
> @Thaodan
>> OK that's disappointing. I don't think it's good to reinvent the wheel
> here.
>>
>> In any case I think there is a reason that most projects end of wrapping
> libnm, i.e. to reduce the overhead on maintaining their NetworkManager
> client API.
>>
>> The NetworkManager documentation recommends to use LibNM for more
> complex cases such as this one:
> https://www.networkmanager.dev/docs/libnm/latest/usage.html
>
>> On another does his have to be added as internal module? Maybe it's
>> better if this lives as an external/separately packaged plugin.
>
> Eventually a lot of the project should probably be split into
> separately installable modules, however that requires some work I
> haven't had any time to do. If it *were* more modular I wouldn't have
> a problem with an nm dep for nm, though my general policy against KDE
> deps remains due to kde/distro release cycles, some distros
> effectively forcing that pulling in one kde dep pulls in all of kde.
>

OK thank you for explaining. Minor comment: I don't think this kind of
issue exist these days anymore at least not since the KF5 days.
I'm not aware of any distributions that have such bad packaging
practices left. There are a few non-KDE Qt programs who have picked up
KDE Frameworks dependencies without those issues. I.e. for example
LxQt.

Anyway thanks for your work.

src/network/wifi.hpp
outfoxxed: Yeah I'm not really expecting more than a connect/disconnect button for broader definitions of "network" like wired connections, vpns, tunnels, and such

src/network/nm/backend.cpp
outfoxxed: Yes, I just remembered qScopeGuard exists as well, which should work for cases like this too.

cpwrs:
@outfoxxed
Thanks for the reviews! It's my first time working with Qt or DBus so I really appreciate the feedback and opportunity.

> ```
> WARN quickshell.network.networkmanager: Sent removal signal for "/org/freedesktop/NetworkManager/Devices/11" which is not registered.
> ```

This was true but not useful. Only warns on device types we support now. Also fixed a bug that was setting security to Unknown when there was a better match.

> There are a couple more options id like to see in network configuration (e.g. per network autoconnect) but we can leave those till this is merged.

I have a big list of stuff that's trivial to add to NM, and a good idea of how IWD will look.
I can make an issue with some TODOs.

> Also have you tested this on a non-open unknown wifi network? It doesn't seem to be able to request a password, at least with the tester.

Yep! The device waits at NMDeviceState.NeedAuth when there's an NM secret agent registered. I've been using nm-applet to test. A few security types have static passwords that can be provided without an agent - adding this to the list.

outfoxxed:
I've made some changes for docs and names that make it a little clearer in my opinion, as well as added connection state on a per network level and a disconnect call on a per network level. It seems like ethernet should be pretty easy from here, but planning to merge without once CI passes

outfoxxed:
Merged! Thanks for dealing with this for so long

PR #114, opened 2025-07-14
core/desktopentry: add file system watcher for desktop entry directories

bbedward:
DesktopEntries only scans one time~~, and doesnt expose scanning functions to QML. This adds a `rescan` function, a signal for applications changed. It just allows a nicer launcher that doesn't get stale.~~

This adds:
- `DesktopUtils` which had a common function for locating `.desktop` files
- `DesktopEntryMonitor` which implements [QFileSystemWatcher](https://doc.qt.io/qt-6/qfilesystemwatcher.html)
  - Watches the desktop entry target directories
- Connect monitor to `DesktopEntryManager` and add a new signal `applicationsChanged`
  - On directory updated, it re-fetches the entries and emits the signal

### DIFF
--- a/src/core/desktopentry.cpp
+++ b/src/core/desktopentry.cpp
@@ -1,22 +1,27 @@
 #include "desktopentry.hpp"
 #include <algorithm>
+#include <utility>
 
 #include <qcontainerfwd.h>
 #include <qdebug.h>
 #include <qdir.h>
+#include <qfile.h>
 #include <qfileinfo.h>
-#include <qhash.h>
-#include <qlist.h>
 #include <qlogging.h>
 #include <qloggingcategory.h>
 #include <qnamespace.h>
 #include <qobject.h>
+#include <qobjectdefs.h>
 #include <qpair.h>
-#include <qstringview.h>
+#include <qproperty.h>
+#include <qscopeguard.h>
 #include <qtenvironmentvariables.h>
+#include <qthreadpool.h>
+#include <qtmetamacros.h>
 #include <ranges>
 
 #include "../io/processcore.hpp"
+#include "desktopentrymonitor.hpp"
 #include "logcat.hpp"
 #include "model.hpp"
 #include "qmlglobal.hpp"
@@ -87,57 +92,60 @@ struct Locale {
 QDebug operator<<(QDebug debug, const Locale& locale) {
 	auto saver = QDebugStateSaver(debug);
 	debug.nospace() << "Locale(language=" << locale.language << ", territory=" << locale.territory
-	                << ", modifier" << locale.modifier << ')';
+	                << ", modifier=" << locale.modifier << ')';
 
 	return debug;
 }
 
-void DesktopEntry::parseEntry(const QString& text) {
+ParsedDesktopEntryData DesktopEntry::parseText(const QString& id, const QString& text) {
+	ParsedDesktopEntryData data;
+	data.id = id;
 	const auto& system = Locale::system();
 
 	auto groupName = QString();
 	auto entries = QHash<QString, QPair<Locale, QString>>();
 
-	auto finishCategory = [this, &groupName, &entries]() {
+	auto finishCategory = [&data, &groupName, &entries]() {
 		if (groupName == "Desktop Entry") {
-			if (entries["Type"].second != "Application") return;
-			if (entries.contains("Hidden") && entries["Hidden"].second == "true") return;
+			if (entries.value("Type").second != "Application") return;
+			if (entries.value("Hidden").second == "true") return;
 
 			for (const auto& [key, pair]: entries.asKeyValueRange()) {
 				auto& [_, value] = pair;
-				this->mEntries.insert(key, value);
+				data.entries.insert(key, value);
 
-				if (key == "Name") this->mName = value;
-				else if (key == "GenericName") this->mGenericName = value;
-				else if (key == "StartupWMClass") this->mStartupClass = value;
-				else if (key == "NoDisplay") this->mNoDisplay = value == "true";
-				else if (key == "Comment") this->mComment = value;
-				else if (key == "Icon") this->mIcon = value;
+				if (key == "Name") data.name = value;
+				else if (key == "GenericName") data.genericName = value;
+				else if (key == "StartupWMClass") data.startupClass = value;
+				else if (key == "NoDisplay") data.noDisplay = value == "true";
+				else if (key == "Comment") data.comment = value;
+				else if (key == "Icon") data.icon = value;
 				else if (key == "Exec") {
-					this->mExecString = value;
-					this->mCommand = DesktopEntry::parseExecString(value);
-				} else if (key == "Path") this->mWorkingDirectory = value;
-				else if (key == "Terminal") this->mTerminal = value == "true";
-				else if (key == "Categories") this->mCategories = value.split(u';', Qt::SkipEmptyParts);
-				else if (key == "Keywords") this->mKeywords = value.split(u';', Qt::SkipEmptyParts);
+					data.execString = value;
+					data.command = DesktopEntry::parseExecString(value);
+				} else if (key == "Path") data.workingDirectory = value;
+				else if (key == "Terminal") data.terminal = value == "true";
+				else if (key == "Categories") data.categories = value.split(u';', Qt::SkipEmptyParts);
+				else if (key == "Keywords") data.keywords = value.split(u';', Qt::SkipEmptyParts);
 			}
 		} else if (groupName.startsWith("Desktop Action ")) {
 			auto actionName = groupName.sliced(16);
-			auto* action = new DesktopAction(actionName, this);
+			DesktopActionData action;
+			action.id = actionName;
 
 			for (const auto& [key, pair]: entries.asKeyValueRange()) {
 				const auto& [_, value] = pair;
-				action->mEntries.insert(key, value);
+				action.entries.insert(key, value);
 
-				if (key == "Name") action->mName = value;
-				else if (key == "Icon") action->mIcon = value;
+				if (key == "Name") action.name = value;
+				else if (key == "Icon") action.icon = value;
 				else if (key == "Exec") {
-					action->mExecString = value;
-					action->mCommand = DesktopEntry::parseExecString(value);
+					action.execString = value;
+					action.command = DesktopEntry::parseExecString(value);
 				}
 			}
 
-			this->mActions.insert(actionName, action);
+			data.actions.insert(actionName, action);
 		}
 
 		entries.clear();
@@ -183,14 +191,62 @@ void DesktopEntry::parseEntry(const QString& text) {
 	}
 
 	finishCategory();
+	return data;
+}
+
+void DesktopEntry::updateState(const ParsedDesktopEntryData& newState) {
+	Qt::beginPropertyUpdateGroup();
+	this->bName = newState.name;
+	this->bGenericName = newState.genericName;
+	this->bStartupClass = newState.startupClass;
+	this->bNoDisplay = newState.noDisplay;
+	this->bComment = newState.comment;
+	this->bIcon = newState.icon;
+	this->bExecString = newState.execString;
+	this->bCommand = newState.command;
+	this->bWorkingDirectory = newState.workingDirectory;
+	this->bRunInTerminal = newState.terminal;
+	this->bCategories = newState.categories;
+	this->bKeywords = newState.keywords;
+	Qt::endPropertyUpdateGroup();
+
+	this->state = newState;
+	this->updateActions(newState.actions);
+}
+
+void DesktopEntry::updateActions(const QHash<QString, DesktopActionData>& newActions) {
+	auto old = this->mActions;
+
+	for (const auto& [key, d]: newActions.asKeyValueRange()) {
+		DesktopAction* act = nullptr;
+		if (auto found = old.find(key); found != old.end()) {
+			act = found.value();
+			old.erase(found);
+		} else {
+			act = new DesktopAction(d.id, this);
+			this->mActions.insert(key, act);
+		}
+
+		Qt::beginPropertyUpdateGroup();
+		act->bName = d.name;
+		act->bIcon = d.icon;
+		act->bExecString = d.execString;
+		act->bCommand = d.command;
+		Qt::endPropertyUpdateGroup();
+
+		act->mEntries = d.entries;
+	}
+
+	for (auto* leftover: old) {
+		leftover->deleteLater();
+	}
 }
 
 void DesktopEntry::execute() const {
-	DesktopEntry::doExec(this->mCommand, this->mWorkingDirectory);
+	DesktopEntry::doExec(this->bCommand.value(), this->bWorkingDirectory.value());
 }
 
-bool DesktopEntry::isValid() const { return !this->mName.isEmpty(); }
-bool DesktopEntry::noDisplay() const { return this->mNoDisplay; }
+bool DesktopEntry::isValid() const { return !this->bName.value().isEmpty(); }
 
 QVector<DesktopAction*> DesktopEntry::actions() const { return this->mActions.values(); }
 
@@ -266,59 +322,44 @@ void DesktopEntry::doExec(const QList<QString>& execString, const QString& worki
 }
 
 void DesktopAction::execute() const {
-	DesktopEntry::doExec(this->mCommand, this->entry->mWorkingDirectory);
+	DesktopEntry::doExec(this->bCommand.value(), this->entry->bWorkingDirectory.value());
 }
 
-DesktopEntryManager::DesktopEntryManager() {
-	this->scanDesktopEntries();
-	this->populateApplications();
+DesktopEntryScanner::DesktopEntryScanner(DesktopEntryManager* manager): manager(manager) {
+	this->setAutoDelete(true);
 }
 
-void DesktopEntryManager::scanDesktopEntries() {
-	QList<QString> dataPaths;
+void DesktopEntryScanner::run() {
+	const auto& desktopPaths = DesktopEntryManager::desktopPaths();
+	auto scanResults = QList<ParsedDesktopEntryData>();
 
-	if (qEnvironmentVariableIsSet("XDG_DATA_HOME")) {
-		dataPaths.push_back(qEnvironmentVariable("XDG_DATA_HOME"));
-	} else if (qEnvironmentVariableIsSet("HOME")) {
-		dataPaths.push_back(qEnvironmentVariable("HOME") + "/.local/share");
+	for (const auto& path: desktopPaths | std::views::reverse) {
+		auto file = QFileInfo(path);
+		if (!file.isDir()) continue;
+
+		this->scanDirectory(QDir(path), QString(), scanResults);
 	}
 
-	if (qEnvironmentVariableIsSet("XDG_DATA_DIRS")) {
-		auto var = qEnvironmentVariable("XDG_DATA_DIRS");
-		dataPaths += var.split(u':', Qt::SkipEmptyParts);
-	} else {
-		dataPaths.push_back("/usr/local/share");
-		dataPaths.push_back("/usr/share");
-	}
-
-	qCDebug(logDesktopEntry) << "Creating desktop entry scanners";
-
-	for (auto& path: std::ranges::reverse_view(dataPaths)) {
-		auto p = QDir(path).filePath("applications");
-		auto file = QFileInfo(p);
-
-		if (!file.isDir()) {
-			qCDebug(logDesktopEntry) << "Not scanning path" << p << "as it is not a directory";
-			continue;
-		}
-
-		qCDebug(logDesktopEntry) << "Scanning path" << p;
-		this->scanPath(p);
-	}
+	QMetaObject::invokeMethod(
+	    this->manager,
+	    "onScanCompleted",
+	    Qt::QueuedConnection,
+	    Q_ARG(QList<ParsedDesktopEntryData>, scanResults)
+	);
 }
 
-void DesktopEntryManager::populateApplications() {
-	for (auto& entry: this->desktopEntries.values()) {
-		if (!entry->noDisplay()) this->mApplications.insertObject(entry);
-	}
-}
+void DesktopEntryScanner::scanDirectory(
+    const QDir& dir,
+    const QString& idPrefix,
+    QList<ParsedDesktopEntryData>& entries
+) {
+	auto dirEntries = dir.entryInfoList(QDir::Dirs | QDir::Files | QDir::NoDotAndDotDot);
 
-void DesktopEntryManager::scanPath(const QDir& dir, const QString& prefix) {
-	auto entries = dir.entryInfoList(QDir::Dirs | QDir::Files | QDir::NoDotAndDotDot);
-
-	for (auto& entry: entries) {
-		if (entry.isDir()) this->scanPath(entry.absoluteFilePath(), prefix + dir.dirName() + "-");
-		else if (entry.isFile()) {
+	for (auto& entry: dirEntries) {
+		if (entry.isDir()) {
+			auto subdirPrefix = idPrefix.isEmpty() ? entry.fileName() : idPrefix + '-' + entry.fileName();
+			this->scanDirectory(QDir(entry.absoluteFilePath()), subdirPrefix, entries);
+		} else if (entry.isFile()) {
 			auto path = entry.filePath();
 			if (!path.endsWith(".desktop")) {
 				qCDebug(logDesktopEntry) << "Skipping file" << path << "as it has no .desktop extension";
@@ -331,44 +372,40 @@ void DesktopEntryManager::scanPath(const QDir& dir, const QString& prefix) {
 				continue;
 			}
 
-			auto id = prefix + entry.fileName().sliced(0, entry.fileName().length() - 8);
-			auto lowerId = id.toLower();
+			auto basename = QFileInfo(entry.fileName()).completeBaseName();
+			auto id = idPrefix.isEmpty() ? basename : idPrefix + '-' + basename;
+			auto content = QString::fromUtf8(file.readAll());
 
-			auto text = QString::fromUtf8(file.readAll());
-			auto* dentry = new DesktopEntry(id, this);
-			dentry->parseEntry(text);
-
-			if (!dentry->isValid()) {
-				qCDebug(logDesktopEntry) << "Skipping desktop entry" << path;
-				delete dentry;
-				continue;
-			}
-
-			qCDebug(logDesktopEntry) << "Found desktop entry" << id << "at" << path;
-
-			auto conflictingId = this->desktopEntries.contains(id);
-
-			if (conflictingId) {
-				qCDebug(logDesktopEntry) << "Replacing old entry for" << id;
-				delete this->desktopEntries.value(id);
-				this->desktopEntries.remove(id);
-				this->lowercaseDesktopEntries.remove(lowerId);
-			}
-
-			this->desktopEntries.insert(id, dentry);
-
-			if (this->lowercaseDesktopEntries.contains(lowerId)) {
-				qCInfo(logDesktopEntry).nospace()
-				    << "Multiple desktop entries have the same lowercased id " << lowerId
-				    << ". This can cause ambiguity when byId requests are not made with the correct case "
-				       "already.";
-
-				this->lowercaseDesktopEntries.remove(lowerId);
-			}
-
-			this->lowercaseDesktopEntries.insert(lowerId, dentry);
+			auto data = DesktopEntry::parseText(id, content);
+			entries.append(std::move(data));
 		}
 	}
+}
+
+DesktopEntryManager::DesktopEntryManager(): monitor(new DesktopEntryMonitor(this)) {
+	QObject::connect(
+	    this->monitor,
+	    &DesktopEntryMonitor::desktopEntriesChanged,
+	    this,
+	    &DesktopEntryManager::handleFileChanges
+	);
+
+	DesktopEntryScanner(this).run();
+}
+
+void DesktopEntryManager::scanDesktopEntries() {
+	qCDebug(logDesktopEntry) << "Starting desktop entry scan";
+
+	if (this->scanInProgress) {
+		qCDebug(logDesktopEntry) << "Scan already in progress, queuing another scan";
+		this->scanQueued = true;
+		return;
+	}
+
+	this->scanInProgress = true;
+	this->scanQueued = false;
+	auto* scanner = new DesktopEntryScanner(this);
+	QThreadPool::globalInstance()->start(scanner);
 }
 
 DesktopEntryManager* DesktopEntryManager::instance() {
@@ -391,14 +428,14 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
 
 	auto list = this->desktopEntries.values();
 
-	auto iter = std::ranges::find_if(list, [&](const DesktopEntry* entry) {
-		return name == entry->mStartupClass;
+	auto iter = std::ranges::find_if(list, [&](DesktopEntry* entry) {
+		return name == entry->bStartupClass.value();
 	});
 
 	if (iter != list.end()) return *iter;
 
-	iter = std::ranges::find_if(list, [&](const DesktopEntry* entry) {
-		return name.toLower() == entry->mStartupClass.toLower();
+	iter = std::ranges::find_if(list, [&](DesktopEntry* entry) {
+		return name.toLower() == entry->bStartupClass.value().toLower();
 	});
 
 	if (iter != list.end()) return *iter;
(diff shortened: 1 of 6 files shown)


outfoxxed:
I'm generally conflicted on this one because I prefer state change notifications to manually triggered scanning, which often leads to polling.

~~As I'm aware, similar GTK based tools usually listen to a signal provided by glib/gio for this. I've been planning to check what that does and implement it here if feasible, as an alternative to re-scanning. If you're interested in figuring it out, https://github.com/GNOME/glib/blob/6e2a31f97a345e0a6b57a0c5c9f9b7514d61b755/gio/gappinfo.c#L2010 is probably a good place to start.~~

Looks like its just a simple directory watch, but it likely has to be recursive on every potential desktop entry dir, otherwise it's unlikely to handle nix generation switches and such.

This is the general direction I'd like to take it in. No pressure on you to implement it if you don't want to.

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+		}
+	} else {
+		// File was deleted
+		this->watcher->removePath(path);
+		this->fileTimestamps.remove(path);
+		this->watchedFiles.remove(path);
+		this->queueChange(ChangeEvent::Removed, path);
+		qCDebug(logDesktopMonitor) << "Desktop file removed:" << path;
+	}
+}
```
outfoxxed: Timestamps cannot be relied on. We shouldn't track or use them.

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+		if (QDir(path).exists()) {
+			qCDebug(logDesktopMonitor) << "Monitoring desktop entry path:" << path;
+			this->addDirectoryHierarchy(path);
+		}
+	}
+}
+
+void DesktopEntryMonitor::addDirectoryHierarchy(const QString& dirPath) {
+	this->scanAndWatch(dirPath);
+}
```
outfoxxed: why?

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+namespace {
+QS_LOGGING_CATEGORY(logDesktopMonitor, "quickshell.desktopentrymonitor", QtWarningMsg);
+}
+
+DesktopEntryMonitor::DesktopEntryMonitor(QObject* parent): QObject(parent) {
+	this->watcher = new QFileSystemWatcher(this);
+	this->debounceTimer = new QTimer(this);
+	this->debounceTimer->setSingleShot(true);
+	this->debounceTimer->setInterval(500); // 500ms debounce
```
outfoxxed: 500ms is unreasonably long for a debounce

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+	    this,
+	    &DesktopEntryMonitor::onDirectoryChanged
+	);
+	connect(
+	    this->watcher,
+	    &QFileSystemWatcher::fileChanged,
+	    this,
+	    &DesktopEntryMonitor::onFileChanged
+	);
+	connect(this->debounceTimer, &QTimer::timeout, this, &DesktopEntryMonitor::processChanges);
```
outfoxxed: QObject::connect

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+}
+
+void DesktopEntryMonitor::addDirectoryHierarchy(const QString& dirPath) {
+	this->scanAndWatch(dirPath);
+}
+
+void DesktopEntryMonitor::scanAndWatch(const QString& dirPath) {
+	QDir dir(dirPath);
```
outfoxxed: disallowed style, use auto

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+			this->addDirectoryHierarchy(path);
+		}
+	}
+}
+
+void DesktopEntryMonitor::addDirectoryHierarchy(const QString& dirPath) {
+	this->scanAndWatch(dirPath);
+}
+
+void DesktopEntryMonitor::scanAndWatch(const QString& dirPath) {
```
outfoxxed, line 54: scanAndWatch + all the various change handlers share a lot of the code. consider initialization to be a diff from an empty set of entries and share the code.

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+		qCDebug(logDesktopMonitor) << "Directory no longer exists, cleaning up:" << path;
+		this->watcher->removePath(path);
+
+		// Remove all watched files from this directory and its subdirectories
+		for (auto it = this->watchedFiles.begin(); it != this->watchedFiles.end();) {
+			const QString& watchedFile = *it;
+			if (watchedFile.startsWith(path + "/") || watchedFile == path) {
+				this->watcher->removePath(watchedFile);
+				this->fileTimestamps.remove(watchedFile);
+				this->queueChange(ChangeEvent::Removed, watchedFile);
```
outfoxxed: The debounce is not used in a logical way. If you are going to debounce, it should guard re-scanning directories, not model updates.

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+	QDir dir(path);
+
+	// Check if directory still exists - handle removal/unmounting
+	if (!dir.exists()) {
+		qCDebug(logDesktopMonitor) << "Directory no longer exists, cleaning up:" << path;
+		this->watcher->removePath(path);
+
+		// Remove all watched files from this directory and its subdirectories
+		for (auto it = this->watchedFiles.begin(); it != this->watchedFiles.end();) {
+			const QString& watchedFile = *it;
```
outfoxxed: Why are we tracking watched files, the fs watcher already does that

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,188 @@
...
+				qCDebug(logDesktopMonitor) << "Removed file due to directory deletion:" << watchedFile;
+				it = this->watchedFiles.erase(it);
+			} else {
+				++it;
+			}
+		}
+		return;
+	}
+
+	QList<QString> currentFiles = dir.entryList({"*.desktop"}, QDir::Files);
```
outfoxxed: auto

outfoxxed (changes requested):
I see you've marked it as WIP. Partial review

src/core/desktopentrymonitor.cpp
bbedward: Stale leftover, I was initially watching .desktop files but decided its better to just watch the directories and re-populate on events.

bbedward:
@outfoxxed  I addressed the comments, I had it initially watching .desktop files and then decided against it, to just watch the directories and rescan when there's a change instead. 

Sorry for the confusion, not wip anymore - ready for review

purian23:
Would there be anything holding this one up for primetime? It will be quite useful.

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// Queue a change to trigger full rescan of all desktop paths
+	this->queueChange(ChangeEvent::Modified, path);
+}
+
+void DesktopEntryMonitor::queueChange(ChangeEvent event, const QString& path) {
+	this->pendingChanges.insert(path, event);
+	this->debounceTimer->start();
+}
```
outfoxxed: Instead of pending changes or tracking, if we're going to do a full rescan anyway then this should do nothing more than set a flag, and potentially bump the debounce timer for a rescan. The timer should be no more than 100ms

src/core/desktopentry.cpp
```diff
@@ -386,7 +383,77 @@ DesktopEntry* DesktopEntryManager::byId(const QString& id) {
...
+	this->mApplications.diffUpdate(newApplications);
+
+	for (auto* e: oldEntries) {
+		if (!this->desktopEntries.contains(e->mId)) {
+			e->deleteLater();
+		}
+	}
+
+	emit applicationsChanged();
+}
```
outfoxxed: Same comment as directory changed

src/core/desktoputils.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <qstringlist.h>
+
+namespace DesktopUtils {
+QList<QString> getDesktopDirectories();
```
outfoxxed: I am not sure why this was extracted into a new file

outfoxxed (changes requested):
Additional notes:
- Please don't add useless comments
- Please do scanning triggered by a change in another thread via QRunnable. The desktop entry scanner takes long enough to run that it may have a noticeable stutter on some systems without it.

src/core/desktopentrymonitor.cpp
bbedward: replaced with `rescanPending` flag

src/core/desktopentry.cpp
bbedward: simply sets flag and starts scan now.

src/core/desktoputils.hpp
bbedward: Moved to `desktopentrymonitor.cpp`

bbedward:
Thanks for taking another look, I implemented `DesktopEntryScanner` QRunnable to handle the rescan in the background and addressed other comments.

src/core/desktopentry.cpp
```diff
@@ -269,56 +267,36 @@ void DesktopAction::execute() const {
...
-
-		qCDebug(logDesktopEntry) << "Scanning path" << p;
-		this->scanPath(p);
-	}
+	QMetaObject::invokeMethod(
+	    this->manager,
+	    "onScanCompleted",
+	    Qt::QueuedConnection,
+	    Q_ARG(DesktopEntryScanResults, scanResults)
+	);
```
outfoxxed: Use the method pointer variant, e.g.
https://github.com/quickshell-mirror/quickshell/blob/f90bef2d994c88f075dbc2fcd81140e160351328/src/core/logging.cpp#L263-L267

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	for (const auto& dir: dataDirs.split(':', Qt::SkipEmptyParts)) {
+		if (!dir.isEmpty()) {
+			dataPaths.append(dir + "/applications");
+		}
+	}
+
+	this->desktopPaths = dataPaths;
+}
```
outfoxxed: This should be stored in a global static, as a function exposed on DesktopEntryManager.
e.g.
```cpp
const QList<QString>& DesktopEntryManager::desktopPaths() {
  static auto paths = []() {
    // ...
  }();

  return paths;
}
```

src/core/desktopentry.cpp
```diff
@@ -407,7 +392,112 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+		auto relInfo = QFileInfo(relativePath);
+
+		auto dirPath = relInfo.path();
+		if (dirPath != ".") {
+			id = dirPath.replace('/', '-') + '-' + relInfo.completeBaseName();
+		}
+	}
+
+	return id;
+}
```
outfoxxed: This should be static

src/core/desktopentry.cpp
```diff
@@ -331,44 +309,51 @@ void DesktopEntryManager::scanPath(const QDir& dir, const QString& prefix) {
...
-			dentry->parseEntry(text);
-
-			if (!dentry->isValid()) {
-				qCDebug(logDesktopEntry) << "Skipping desktop entry" << path;
-				delete dentry;
-				continue;
-			}
+			ParsedDesktopEntry parsed;
+			parsed.id = id;
+			parsed.content = std::move(content);
```
outfoxxed: Desktop entry content should be parsed off-thread. Use bindable properties to update the fields of desktopentry/desktopaction.

src/core/desktopentry.cpp
```diff
@@ -407,7 +392,112 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+
+	for (auto* e: oldEntries) {
+		if (!this->desktopEntries.contains(e->mId)) {
+			e->deleteLater();
+		}
+	}
+	oldEntries.clear();
+
+	emit applicationsChanged();
+}
```
outfoxxed: DiffUpdate will not do anything useful when the values are different. This has to be handled manually. See above note

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,113 @@
...
+			this->watcher->addPath(subdirPath);
+		}
+	}
+}
+
+void DesktopEntryMonitor::onDirectoryChanged(const QString& path) {
+	auto dir = QDir(path);
+
+	if (!dir.exists()) {
+		this->watcher->removePath(path);
```
outfoxxed: This is redundant

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,113 @@
...
+	if (!dir.exists()) return;
+
+	if (!this->watcher->directories().contains(dirPath)) {
+		this->watcher->addPath(dirPath);
+	}
+
+	auto subdirs = dir.entryInfoList(QDir::Dirs | QDir::NoDotAndDotDot);
+	for (const auto& subdir: subdirs) {
+		auto subdirPath = subdir.absoluteFilePath();
+		if (!this->watcher->directories().contains(subdirPath)) {
```
outfoxxed: This is redundant

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+void DesktopEntryMonitor::scanAndWatch(const QString& dirPath) {
+	auto dir = QDir(dirPath);
+	if (!dir.exists()) return;
+
+	if (!this->watcher->directories().contains(dirPath)) {
+		this->watcher->addPath(dirPath);
+	}
```
outfoxxed: as is this if statement

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+	auto watchedPaths = this->watcher->directories();
+	for (const auto& path: watchedPaths) {
+		if (!QDir(path).exists()) {
+			this->watcher->removePath(path);
+		}
+	}
+
+	emit desktopEntriesChanged();
+}
```
outfoxxed: Old paths shouldn't need to be removed as the watcher handles that implicitly

src/core/desktopentry.cpp
bbedward: Done

src/core/desktopentry.cpp
bbedward: Doing the parsing in DesktopEntryScanner thread now, and made properties bindable, update them in onScanCompleted.

src/core/desktopentry.cpp
bbedward: Re-using existing entries* when IDs match now, deleting leftovers

src/core/desktopentry.hpp
```diff
@@ -148,6 +222,23 @@ class DesktopAction: public QObject {
 	friend class DesktopEntry;
 };
 
+class DesktopEntryManager;
+
+using DesktopEntryScanResults = QList<ParsedDesktopEntryData>;
+Q_DECLARE_METATYPE(ParsedDesktopEntryData)
+Q_DECLARE_METATYPE(DesktopEntryScanResults)
```
outfoxxed: why is this typdef'd or a metatype

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,66 @@
...
+	this->startMonitoring();
+}
+
+void DesktopEntryMonitor::startMonitoring() {
+	for (const auto& path: DesktopEntryManager::desktopPaths()) {
+		if (QDir(path).exists()) {
+			this->scanAndWatch(path);
+		}
+	}
+}
```
outfoxxed: Not only the given path, but every parent must be watched to detect symlinked dirs being swapped out.

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,66 @@
...
+	auto subdirs = dir.entryInfoList(QDir::Dirs | QDir::NoDotAndDotDot);
+	for (const auto& subdir: subdirs) {
+		this->watcher->addPath(subdir.absoluteFilePath());
+	}
+}
+
+void DesktopEntryMonitor::onDirectoryChanged(const QString& path) {
+	this->scanAndWatch(path);
+
+	if (!this->rescanPending) {
```
outfoxxed: Ensuring this is false is unhelpful for the purposes of a debounce. We want to reset the timer if something else changes. It doesn't look like rescanPending is even needed, looking at its usages.

src/core/desktopentrymonitor.hpp
```diff
@@ -0,0 +1,29 @@
...
+private slots:
+	void onDirectoryChanged(const QString& path);
+	void processChanges();
+
+private:
+	void startMonitoring();
+	void scanAndWatch(const QString& dirPath);
+
+	QFileSystemWatcher* watcher;
+	QTimer* debounceTimer;
```
outfoxxed: These don't need to be pointers, they can just be inlined.

src/core/desktopentry.cpp
```diff
@@ -183,20 +188,55 @@ void DesktopEntry::parseEntry(const QString& text) {
...
+	this->bExecString = newState.execString;
+	this->bCommand = newState.command;
+	this->bWorkingDirectory = newState.workingDirectory;
+	this->bRunInTerminal = newState.terminal;
+	this->bCategories = newState.categories;
+	this->bKeywords = newState.keywords;
+
+	this->state = newState;
+	this->updateActions(newState.actions);
+}
```
outfoxxed: Use a property update group

src/core/desktopentry.cpp
```diff
@@ -183,20 +188,55 @@ void DesktopEntry::parseEntry(const QString& text) {
...
+
+	for (const auto& [actionName, actionData]: newActions.asKeyValueRange()) {
+		auto* action = new DesktopAction(actionData.id, this);
+		action->mEntries = actionData.entries;
+		action->mName = actionData.name;
+		action->mIcon = actionData.icon;
+		action->mExecString = actionData.execString;
+		action->mCommand = actionData.command;
+		this->mActions.insert(actionName, action);
+	}
```
outfoxxed: This should be handled with a diff update, and bindables in DesktopAction should be used.

src/core/desktopentry.cpp
```diff
@@ -407,7 +425,156 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+		auto dataDirs = qEnvironmentVariable("XDG_DATA_DIRS");
+		if (dataDirs.isEmpty()) {
+			dataDirs = "/usr/local/share:/usr/share";
+		}
+
+		for (const auto& dir: dataDirs.split(':', Qt::SkipEmptyParts)) {
+			if (!dir.isEmpty()) {
+				dataPaths.append(dir + "/applications");
+			}
+		}
```
outfoxxed: Use single line ifs when short

src/core/desktopentry.cpp
```diff
@@ -407,7 +425,156 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+	}
+
+	this->mApplications.diffUpdate(newApplications);
+
+	for (auto* e: oldEntries) {
+		e->deleteLater();
+	}
+
+	emit applicationsChanged();
+}
```
outfoxxed: Single line ifs/fors when short, newlines after },

src/core/desktopentry.cpp
```diff
@@ -407,7 +425,156 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+				dentry->deleteLater();
+			}
+			continue;
+		}
+
+		qCDebug(logDesktopEntry) << "Found desktop entry" << data.id;
+
+		auto lowerId = data.id.toLower();
+
+		auto conflictingId = newEntries.contains(data.id);
```
outfoxxed: No newline

outfoxxed (changes requested):

src/core/desktopentry.hpp
bbedward: Removed by using lambda-based invokeMethod

src/core/desktopentrymonitor.cpp
bbedward: Added addPathAndParents to watch this case

src/core/desktopentrymonitor.cpp
bbedward: Simplified this so it basically just starts the debounce,  no rescanPending flag

src/core/desktopentry.cpp
```diff
@@ -87,64 +86,71 @@ struct Locale {
...
+	auto finishCategory = [&data, &groupName, &entries]() {
 		if (groupName == "Desktop Entry") {
-			if (entries["Type"].second != "Application") return;
-			if (entries.contains("Hidden") && entries["Hidden"].second == "true") return;
+			const auto typeIt = entries.constFind("Type");
+			if (typeIt == entries.cend() || typeIt->second != "Application") return;
+
+			if (const auto hiddenIt = entries.constFind("Hidden");
+			    hiddenIt != entries.cend() && hiddenIt->second == "true")
+				return;
```
outfoxxed: just use .value("Type") etc

src/core/desktopentry.cpp
```diff
@@ -87,64 +86,71 @@ struct Locale {
...
+					data.command = DesktopEntry::parseExecString(value);
+				} else if (key == "Path") data.workingDirectory = value;
+				else if (key == "Terminal") data.terminal = value == "true";
+				else if (key == "Categories") data.categories = value.split(u';', Qt::SkipEmptyParts);
+				else if (key == "Keywords") data.keywords = value.split(u';', Qt::SkipEmptyParts);
 			}
 		} else if (groupName.startsWith("Desktop Action ")) {
 			auto actionName = groupName.sliced(16);
-			auto* action = new DesktopAction(actionName, this);
+			auto action = DesktopActionData();
```
outfoxxed: DesktopActionData action;

src/core/desktopentry.cpp
```diff
@@ -162,7 +168,7 @@ void DesktopEntry::parseEntry(const QString& text) {
 		const auto& value = line.sliced(splitIdx + 1);
 
 		auto localeIdx = key.indexOf('[');
-		Locale locale;
+		auto locale = Locale();
```
outfoxxed: param-less init shouldn't use auto

src/core/desktopentry.cpp
```diff
@@ -183,20 +189,71 @@ void DesktopEntry::parseEntry(const QString& text) {
...
+	this->state = newState;
+	this->updateActions(newState.actions);
+}
+
+void DesktopEntry::updateActions(const QHash<QString, DesktopActionData>& newActions) {
+	auto old = this->mActions;
+
+	for (auto it = newActions.cbegin(); it != newActions.cend(); ++it) {
+		const auto& key = it.key();
+		const auto& d = it.value();
```
outfoxxed: Avoid explicit iterators when not required for manipulation during iteration.

src/core/desktopentry.cpp
```diff
@@ -183,20 +189,71 @@ void DesktopEntry::parseEntry(const QString& text) {
...
+		act->bName = d.name;
+		act->bIcon = d.icon;
+		act->bExecString = d.execString;
+		act->bCommand = d.command;
+		Qt::endPropertyUpdateGroup();
+
+		act->mEntries = d.entries;
+	}
+
+	for (auto* leftover: std::as_const(old)) {
```
outfoxxed: This should implicitly be using const-iterators, and it won't even cause a detach if it isn't.

src/core/desktopentry.cpp
```diff
@@ -183,20 +189,71 @@ void DesktopEntry::parseEntry(const QString& text) {
...
-bool DesktopEntry::noDisplay() const { return this->mNoDisplay; }
+bool DesktopEntry::isValid() const { return !this->bName.value().isEmpty(); }
 
 QVector<DesktopAction*> DesktopEntry::actions() const { return this->mActions.values(); }
 
 QVector<QString> DesktopEntry::parseExecString(const QString& execString) {
-	QVector<QString> arguments;
-	QString currentArgument;
+	auto arguments = QVector<QString>();
+	auto currentArgument = QString();
```
outfoxxed: Please don't do this, same as earlier comment

src/core/desktopentry.cpp
```diff
@@ -259,66 +316,46 @@ QVector<QString> DesktopEntry::parseExecString(const QString& execString) {
...
-	}
-}
-
-void DesktopEntryManager::populateApplications() {
-	for (auto& entry: this->desktopEntries.values()) {
-		if (!entry->noDisplay()) this->mApplications.insertObject(entry);
-	}
+	QMetaObject::invokeMethod(
+	    this->manager,
+	    [mgr = this->manager, results = std::move(scanResults)]() { mgr->onScanCompleted(results); },
```
outfoxxed: Args can be passed to invokeMethod
https://doc.qt.io/qt-6/qmetaobject.html#invokeMethod-8

src/core/desktopentry.cpp
```diff
@@ -331,44 +368,38 @@ void DesktopEntryManager::scanPath(const QDir& dir, const QString& prefix) {
...
-			if (!dentry->isValid()) {
-				qCDebug(logDesktopEntry) << "Skipping desktop entry" << path;
-				delete dentry;
-				continue;
-			}
-
-			qCDebug(logDesktopEntry) << "Found desktop entry" << id << "at" << path;
-
-			auto conflictingId = this->desktopEntries.contains(id);
+			auto id = DesktopEntryManager::extractIdFromPath(entry.absoluteFilePath());
```
outfoxxed: It seems more logical to pass an id prefix string to scanDirectory, then prepend it to the filename sub .desktop than to compute the filename from the full path.

src/core/desktopentry.cpp
```diff
@@ -407,7 +438,138 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+		auto dataHome = qEnvironmentVariable("XDG_DATA_HOME");
+		if (dataHome.isEmpty() && qEnvironmentVariableIsSet("HOME"))
+			dataHome = qEnvironmentVariable("HOME") + "/.local/share";
+		if (!dataHome.isEmpty()) dataPaths.append(dataHome + "/applications");
+
+		auto dataDirs = qEnvironmentVariable("XDG_DATA_DIRS");
+		if (dataDirs.isEmpty()) dataDirs = "/usr/local/share:/usr/share";
+
+		for (const auto& dir: dataDirs.split(':', Qt::SkipEmptyParts))
+			dataPaths.append(dir + "/applications");
```
outfoxxed: My bad for being unclear. Single line ifs/fors *only* if it doesn't cause a line break.

src/core/desktopentry.cpp
```diff
@@ -407,7 +438,138 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+
+	for (auto* e: oldEntries) e->deleteLater();
+
+	emit applicationsChanged();
+}
+
+DesktopEntries::DesktopEntries() {
+	auto* mgr = DesktopEntryManager::instance();
+	QObject::connect(
+	    mgr,
```
outfoxxed: inline mgr

src/core/desktopentry.hpp
```diff
@@ -6,35 +6,73 @@
...
 	/// Name of the icon associated with this application. May be empty.
-	Q_PROPERTY(QString icon MEMBER mIcon CONSTANT);
+	Q_PROPERTY(QString icon READ icon NOTIFY iconChanged BINDABLE bindableIcon);
 	/// The raw `Exec` string from the desktop entry.
 	///
 	/// > [!WARNING] This cannot be reliably run as a command. See @@command for one you can run.
-	Q_PROPERTY(QString execString MEMBER mExecString CONSTANT);
+	Q_PROPERTY(
+	    QString execString READ execString NOTIFY execStringChanged BINDABLE bindableExecString
+	);
```
outfoxxed: Line breaks in Q_PROPERTY definitions will break the documentation generator. Use // clang-format off.

Additionally, I am aware this isn't documented anywhere yet, but for new code I require properties using BINDABLE to use `READ default` and `WRITE default` if either READ or WRITE is required, or drop the BINDABLE. The QML engine may bypass the getters/setters if BINDABLE is provided so they have to be semantically equivalent anyway.

src/core/desktopentry.hpp
```diff
@@ -73,30 +123,111 @@ class DesktopEntry: public QObject {
...
+	    QVector<QString>,
+	    bCategories,
+	    &DesktopEntry::categoriesChanged
+	);
+	Q_OBJECT_BINDABLE_PROPERTY(
+	    DesktopEntry,
+	    QVector<QString>,
+	    bKeywords,
+	    &DesktopEntry::keywordsChanged
+	);
```
outfoxxed: Please also keep these on single lines. Its easier to make an editor window wider than to read them spread out like this.

src/core/desktopentry.hpp
```diff
@@ -73,30 +123,111 @@ class DesktopEntry: public QObject {
...
+	[[nodiscard]] QString startupClass() const { return this->bStartupClass; }
+	[[nodiscard]] bool noDisplay() const { return this->bNoDisplay; }
+	[[nodiscard]] QString comment() const { return this->bComment; }
+	[[nodiscard]] QString icon() const { return this->bIcon; }
+	[[nodiscard]] QString execString() const { return this->bExecString; }
+	[[nodiscard]] QVector<QString> command() const { return this->bCommand; }
+	[[nodiscard]] QString workingDirectory() const { return this->bWorkingDirectory; }
+	[[nodiscard]] bool runInTerminal() const { return this->bRunInTerminal; }
+	[[nodiscard]] QVector<QString> categories() const { return this->bCategories; }
+	[[nodiscard]] QVector<QString> keywords() const { return this->bKeywords; }
```
outfoxxed: Drop all these in favor of `READ default`

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <qdiriterator.h>
+#include <qfileinfo.h>
+#include <qloggingcategory.h>
+
+#include "desktopentry.hpp"
+
+static void addPathAndParents(QFileSystemWatcher& watcher, const QString& path) {
+	watcher.addPath(path);
+
+	QString p = QFileInfo(path).absolutePath();
```
outfoxxed: auto unless it returns something other than a QString

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,57 @@
...
+	auto dir = QDir(dirPath);
+	if (!dir.exists()) return;
+
+	this->watcher.addPath(dirPath);
+
+	auto subdirs = dir.entryInfoList(QDir::Dirs | QDir::NoDotAndDotDot | QDir::NoSymLinks);
+	for (const auto& subdir: subdirs) this->watcher.addPath(subdir.absoluteFilePath());
+}
+
+void DesktopEntryMonitor::onDirectoryChanged(const QString&) { this->debounceTimer.start(); }
```
outfoxxed: Unfortunately once we include parents it becomes more complex than this. Ideally we don't want to trigger a rescan if a user creates a file in / or something. However this is acceptable for now because fixing it properly will require rewriting QFileSystemWatcher.

outfoxxed (changes requested):
Sorry this one has required a lot of re-review. Hopefully we're near done.

bbedward:
Think I addressed all of the ones from this round @outfoxxed  hopefully good now  🤞

outfoxxed:
missed the collapsed ones?

bbedward:
> missed the collapsed ones?

yep, just addressed them - hopefully good now for real 🤦

src/core/desktopentry.cpp
```diff
@@ -87,64 +86,68 @@ struct Locale {
...
 	auto entries = QHash<QString, QPair<Locale, QString>>();
 
-	auto finishCategory = [this, &groupName, &entries]() {
+	auto finishCategory = [&data, &groupName, &entries]() {
 		if (groupName == "Desktop Entry") {
-			if (entries["Type"].second != "Application") return;
-			if (entries.contains("Hidden") && entries["Hidden"].second == "true") return;
+			if (entries.value("Type").second != "Application") return;
+
+			if (entries.value("Hidden").second == "true") return;
```
outfoxxed: Why the extra newline?

src/core/desktopentry.cpp
```diff
@@ -87,64 +86,68 @@ struct Locale {
...
+			data.actions.insert(actionName, action);
 		}
 
 		entries.clear();
 	};
 
-	for (auto& line: text.split(u'\n', Qt::SkipEmptyParts)) {
-		if (line.startsWith(u'#')) continue;
+	for (auto& line: text.split(u'\n')) {
+		if (line.isEmpty() || line.startsWith(u'#')) continue;
```
outfoxxed: Why?

src/core/desktopentry.cpp
```diff
@@ -183,14 +186,63 @@ void DesktopEntry::parseEntry(const QString& text) {
...
+	this->state = newState;
+	this->updateActions(newState.actions);
+}
+
+void DesktopEntry::updateActions(const QHash<QString, DesktopActionData>& newActions) {
+	auto old = this->mActions;
+
+	for (const auto& [key, d]: newActions.asKeyValueRange()) {
+
+		DesktopAction* act = nullptr;
```
outfoxxed: newline

src/core/desktopentry.cpp
```diff
@@ -259,66 +311,52 @@ QVector<QString> DesktopEntry::parseExecString(const QString& execString) {
 }
 
 void DesktopEntry::doExec(const QList<QString>& execString, const QString& workingDirectory) {
-	qs::io::process::ProcessContext ctx;
+	auto ctx = qs::io::process::ProcessContext();
```
outfoxxed: fix

src/core/desktopentry.cpp
```diff
@@ -259,66 +311,52 @@ QVector<QString> DesktopEntry::parseExecString(const QString& execString) {
...
-	this->scanDesktopEntries();
-	this->populateApplications();
+DesktopEntryScanner::DesktopEntryScanner(DesktopEntryManager* manager): manager(manager) {
+	this->setAutoDelete(true);
 }
 
-void DesktopEntryManager::scanDesktopEntries() {
-	QList<QString> dataPaths;
+void DesktopEntryScanner::run() {
+	auto desktopPaths = DesktopEntryManager::desktopPaths();
```
outfoxxed: const auto&

src/core/desktopentry.cpp
```diff
@@ -259,66 +311,52 @@ QVector<QString> DesktopEntry::parseExecString(const QString& execString) {
...
+	auto scanResults = QList<ParsedDesktopEntryData>();
 
-	if (qEnvironmentVariableIsSet("XDG_DATA_HOME")) {
-		dataPaths.push_back(qEnvironmentVariable("XDG_DATA_HOME"));
-	} else if (qEnvironmentVariableIsSet("HOME")) {
-		dataPaths.push_back(qEnvironmentVariable("HOME") + "/.local/share");
-	}
+	for (int i = desktopPaths.size() - 1; i >= 0; --i) {
+		const auto& path = desktopPaths.at(i);
```
outfoxxed: Use a reverse view

src/core/desktopentrymonitor.cpp
```diff
@@ -0,0 +1,57 @@
...
+	this->debounceTimer.setSingleShot(true);
+	this->debounceTimer.setInterval(100);
+
+	connect(
+	    &this->watcher,
+	    &QFileSystemWatcher::directoryChanged,
+	    this,
+	    &DesktopEntryMonitor::onDirectoryChanged
+	);
+	connect(&this->debounceTimer, &QTimer::timeout, this, &DesktopEntryMonitor::processChanges);
```
outfoxxed: QObject::

src/core/desktopentry.cpp
```diff
@@ -407,7 +440,115 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
 ObjectModel<DesktopEntry>* DesktopEntryManager::applications() { return &this->mApplications; }
 
-DesktopEntries::DesktopEntries() { DesktopEntryManager::instance(); }
+void DesktopEntryManager::handleFileChanges() {
+	qCDebug(logDesktopEntry) << "Directory change detected, performing full rescan";
+
+	if (this->scanInProgress) {
+		qCDebug(logDesktopEntry) << "Scan already in progress, skipping";
+		return;
+	}
```
outfoxxed: This looks race-prone. Consider the following case:
1. An app is added to a scan dir, we begin a scan
2. Another app is added once we've already ls'd the dir, scanInProgress is set so it does nothing
3. We finish the scan, unset scanInProgress, and don't find the other app

I would recommend a scanQueued flag or equivalent that kicks off another scan when we'd set scanInProgress to false, ref desktopentry.cpp:496

src/core/desktopentry.cpp
```diff
@@ -407,7 +440,115 @@ DesktopEntry* DesktopEntryManager::heuristicLookup(const QString& name) {
...
+	this->desktopEntries = newEntries;
+	this->lowercaseDesktopEntries = newLowercaseEntries;
+
+	auto newApplications = QVector<DesktopEntry*>();
+	for (auto* entry: this->desktopEntries.values())
+		if (!entry->bNoDisplay) newApplications.append(entry);
+
+	this->mApplications.diffUpdate(newApplications);
+
+	for (auto* e: oldEntries) e->deleteLater();
```
outfoxxed: Some things do require a deleteLater to behave well. Is this one of them?

Emit this->applicationsChanged() first so the deletes don't trigger signals in qml though.

outfoxxed (changes requested):
I've received a report that this patch also duplicates the desktop entry list in rare cases, though I didn't see anything that would obviously cause that.
Apologies for the long review turnaround as well, I've been away for the last couple weeks.

src/core/desktopentry.cpp
bbedward: Added scanQueued flag and check

bbedward:
Haven't seen the entries get duplicated myself , addressed the latest round of comments

outfoxxed:
Sorry its taken forever, I've had a lot going on. Anyway I've tested that this works on nix and I'm going to go review it again.

outfoxxed (approved):
I think this is acceptable to merge.

As a note for the future, once QS has its own inotify handler, the rescan trigger should be much more fine grained, but for now it works.

@bbedward Thanks for the work and putting up with my schedule!

src/core/desktopentry.cpp
```diff
@@ -377,6 +377,9 @@ void DesktopEntryScanner::scanDirectory(
 }
 
 DesktopEntryManager::DesktopEntryManager() {
+	qRegisterMetaType<ParsedDesktopEntryData>("ParsedDesktopEntryData");
+	qRegisterMetaType<QList<ParsedDesktopEntryData>>("QList<ParsedDesktopEntryData>");
+
```
outfoxxed: This is not actually required

src/core/desktopentry.hpp
```diff
@@ -45,6 +45,9 @@ struct ParsedDesktopEntryData {
 	QHash<QString, DesktopActionData> actions;
 };
 
+Q_DECLARE_METATYPE(ParsedDesktopEntryData)
+Q_DECLARE_METATYPE(QList<ParsedDesktopEntryData>)
+
```
outfoxxed: Or this

outfoxxed (changes requested):
I've noticed one other breaking change, which is that the initial scan is now non-blocking. While that might be desireable it will also break animations in existing configs.

Since these are trivial fixes I'll just do them now and merge.

outfoxxed (approved):

PR #115, opened 2025-07-14
service/notifications: add inline-reply action support

ipg0:
Added support for _quite-but-not-really-standard_ `inline-reply` action to notifications.
This action is used by apps like Telegram to provide a quick reply text field in the notification UI.
The text is then submitted using the non-standard `org.freedesktop.Notifications.NotificationReplied` D-Bus signal, which is implemented in this PR.

- Added `NotificationServer.inlineReplySupported` option
- Added `Notification.sendInlineReply(const QString& replyText)` method to send the reply
- Added `Notification.hasInlineReply` property to simplify conditional rendering of the text field

### DIFF
--- a/src/services/notifications/notification.cpp
+++ b/src/services/notifications/notification.cpp
@@ -78,6 +78,29 @@ void Notification::close(NotificationCloseReason::Enum reason) {
 	}
 }
 
+void Notification::sendInlineReply(const QString& replyText) {
+	if (!NotificationServer::instance()->support.inlineReply) {
+		qCritical() << "Inline reply support disabled on server";
+		return;
+	}
+
+	if (!this->bHasInlineReply) {
+		qCritical() << "Cannot send reply to notification without inline-reply action";
+		return;
+	}
+
+	if (this->isRetained()) {
+		qCritical() << "Cannot send reply to destroyed notification" << this;
+		return;
+	}
+
+	NotificationServer::instance()->NotificationReplied(this->id(), replyText);
+
+	if (!this->bindableResident().value()) {
+		this->close(NotificationCloseReason::Dismissed);
+	}
+}
+
 void Notification::updateProperties(
     const QString& appName,
     QString appIcon,
@@ -147,17 +170,27 @@ void Notification::updateProperties(
 	this->bImage = imagePath;
 	this->bHints = hints;
 
-	Qt::endPropertyUpdateGroup();
-
 	bool actionsChanged = false;
 	auto deletedActions = QVector<NotificationAction*>();
 
 	if (actions.length() % 2 == 0) {
 		int ai = 0;
 		for (auto i = 0; i != actions.length(); i += 2) {
-			ai = i / 2;
 			const auto& identifier = actions.at(i);
 			const auto& text = actions.at(i + 1);
+
+			if (identifier == "inline-reply" && NotificationServer::instance()->support.inlineReply) {
+				if (this->bHasInlineReply) {
+					qCWarning(logNotifications) << this << '(' << appName << ')'
+					                            << "sent an action set with duplicate inline-reply actions.";
+				} else {
+					this->bHasInlineReply = true;
+					this->bInlineReplyPlaceholder = text;
+				}
+				// skip inserting this action into action list
+				continue;
+			}
+
 			auto* action = ai < this->mActions.length() ? this->mActions.at(ai) : nullptr;
 
 			if (action && identifier == action->identifier()) {
@@ -187,6 +220,8 @@ void Notification::updateProperties(
 		qCWarning(logNotifications) << this << '(' << appName << ')'
 		                            << "sent an action set of an invalid length.";
 	}
+
+	Qt::endPropertyUpdateGroup();
 
 	if (actionsChanged) emit this->actionsChanged();
 
--- a/src/services/notifications/notification.hpp
+++ b/src/services/notifications/notification.hpp
@@ -107,6 +107,12 @@ class Notification
 	///
 	/// This image is often something like a profile picture in instant messaging applications.
 	Q_PROPERTY(QString image READ default NOTIFY imageChanged BINDABLE bindableImage);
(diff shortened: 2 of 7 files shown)


outfoxxed:
Are there any apps other than telegram that use this that I can test against, and any informal spec locations?

ipg0:
@outfoxxed afaik most of them, it's what other notification daemons like swaync do.

Can't test right now, but I think at least Discord should do that too

ipg0:
@outfoxxed afaik most of them, it's what other notification daemons like swaync do.

Can't test right now, but I think at least Discord should do that too

Edit: as per the specs, I found a proposal on the mail archive that pretty much describes what I implemented:

https://www.mail-archive.com/xdg@lists.freedesktop.org/msg08777.html

outfoxxed:
This seems like its implemented widely enough to make sense

src/services/notifications/notification.cpp
```diff
@@ -78,6 +78,28 @@ void Notification::close(NotificationCloseReason::Enum reason) {
...
 
+void Notification::sendInlineReply(const QString& replyText) {
+	if (!NotificationServer::instance()->support.inlineReply) {
+		qCritical() << "Inline reply support disabled on server";
+		return;
+	}
+
+	if (!this->bHasInlineReply) {
+		qCritical() << "Cannot send reply to notification without inline-reply action";
+	}
```
outfoxxed: Return?

src/services/notifications/notification.cpp
```diff
@@ -193,6 +215,15 @@ void Notification::updateProperties(
...
 	}
+
+	auto inlineReplyActionIterator = std::find_if(
+	    this->actions().begin(),
+	    this->actions().end(),
+	    [](const NotificationAction* action) -> bool {
+		    return (action->identifier() == "inline-reply");
+	    }
+	);
+	this->bHasInlineReply = inlineReplyActionIterator != this->actions().end();
```
outfoxxed: The inline reply action, if supported by the server, should be removed from mActions. We should keep a separate property for inline reply placeholder text.

The placeholder / hasInlineReply flag can be discovered while iterating over the actions list, and assigned to bindables following it.

src/services/notifications/notification.hpp
```diff
@@ -124,6 +129,12 @@ class Notification
 	/// explicitly closed by the user.
 	Q_INVOKABLE void dismiss();
 
+	/// Send an inline reply to the notification with an inline reply action.
+	/// > [!WARNING] This method will only run if the notification
+	/// > has an action with identifier "inline-reply"
```
outfoxxed: Since this is user facing, phrase it in terms of exposed properties, e.g. "This method can only be called if @@hasInlineReply is true."

src/services/notifications/qml.cpp
```diff
@@ -115,6 +115,15 @@ void NotificationServerQml::setImageSupported(bool imageSupported) {
 	emit this->imageSupportedChanged();
 }
 
+bool NotificationServerQml::inlineReplySupported() const { return this->support.inlineReply; }
+
+void NotificationServerQml::setInlineReplySupported(bool inlineReplySupported) {
+	if (inlineReplySupported == this->support.inlineReply) return;
+	this->support.inlineReply = inlineReplySupported;
+	this->updateSupported();
+	emit this->imageSupportedChanged();
```
outfoxxed: imageSupportChanged?

src/services/notifications/server.cpp
```diff
@@ -156,6 +156,8 @@ QStringList NotificationServer::GetCapabilities() const {
 
 	if (this->support.image) capabilities += "icon-static";
 
+	if (this->support.inlineReply) capabilities += "inline-reply";
```
outfoxxed: no newline between these please

outfoxxed (changes requested):

ipg0:
@outfoxxed I'll fix the linter error shortly

src/services/notifications/notification.cpp
ipg0: yep, fixing

src/services/notifications/notification.hpp
ipg0: makes sense, fixing

src/services/notifications/qml.cpp
ipg0: fixing

src/services/notifications/server.cpp
ipg0: fixing

src/services/notifications/notification.cpp
ipg0: fixing

ipg0:
@outfoxxed I can't run linting locally bc the env `TIDYFOX` is not defined and I have no idea what it should be.

Otherwise all fixed.

src/services/notifications/notification.hpp
```diff
@@ -112,6 +112,9 @@ class Notification
 	/// A quick reply text field should be displayed and the reply can be sent using @@sendInlineReply().
 	Q_PROPERTY(bool hasInlineReply READ default NOTIFY hasInlineReplyChanged BINDABLE bindableHasInlineReply);
 
+	Q_PROPERTY(QString inlineReplyPlaceholder READ default NOTIFY inlineReplyPlaceholderChanged BINDABLE bindableInlineReplyPlaceholder);
+
+
 	/// All hints sent by the client application as a javascript object.
```
outfoxxed: Missing desc, odd spacing

outfoxxed (changes requested):
> @outfoxxed I can't run linting locally bc the env TIDYFOX is not defined and I have no idea what it should be.

Its a clang-tidy plugin, all it does is enforce thisptr usage. you can copy the command from the justfile and remove the --load flag.

src/services/notifications/notification.hpp
ipg0: fixed

ipg0:
@outfoxxed everything seems ok, fixed the comment

PR #127, opened 2025-07-17
bluetooth: add rssi property

bbedward:
Add rssi to bluetooth property bindings. It's useful to see signal strength of devices during discovery (IE, for sorting by signal strength, ignoring very low signal strength devices)

Because rssi is not available when the adapter is no longer discovering I also added a condition to lower the log level on the property update callback for optional properties, it's not really necessary and can be removed - but prevents some log spam when  turning off discovery (because the rssi properties become unavailable)

### DIFF
--- a/src/bluetooth/device.cpp
+++ b/src/bluetooth/device.cpp
@@ -56,24 +56,23 @@ BluetoothAdapter* BluetoothDevice::adapter() const {
 }
 
 qint32 BluetoothDevice::signalStrength() const {
-	// Convert RSSI (dBm) to a normalized 0-100 scale
-	// Based on practical Bluetooth RSSI ranges:
-	// -30 to -40 dBm = 85-100
-	// -40 to -55 dBm = 65-85
-	// -55 to -65 dBm = 45-65
-	// -65 to -75 dBm = 25-45
-	// -75 to -85 dBm = 10-25
-	// <= -85 dBm     = 0-10
+	if (this->bRssi.value() == 0) return 0;
 
-	auto rssiValue = this->bRssi.value();
-	if (rssiValue == 0) {
-		return 0;
+	auto rssi = std::clamp(static_cast<double>(this->bRssi.value()), -100.0, -30.0);
+
+	if (rssi >= -60.0) {
+		auto percentage = 75.0 + ((rssi + 60.0) / 30.0) * 25.0;
+		return static_cast<qint32>(std::round(percentage));
 	}
-
-	auto rssi = std::max(static_cast<qint16>(-100), std::min(static_cast<qint16>(-30), rssiValue));
-	auto normalized = static_cast<qint32>(((rssi + 100) / 70.0) * 100.0);
-
-	return std::max(0, std::min(100, normalized));
+	if (rssi >= -80.0) {
+		auto percentage = 9.0 + ((rssi + 80.0) / 20.0) * 66.0;
+		return static_cast<qint32>(std::round(percentage));
+	}
+	if (rssi >= -90.0) {
+		auto percentage = 0.0 + ((rssi + 90.0) / 10.0) * 9.0;
+		return static_cast<qint32>(std::round(percentage));
+	}
+	return 0;
 }
 
 void BluetoothDevice::setConnected(bool connected) {
--- a/src/bluetooth/device.cpp
+++ b/src/bluetooth/device.cpp
@@ -1,4 +1,5 @@
 #include "device.hpp"
+#include <algorithm>
 
 #include <qcontainerfwd.h>
 #include <qdbusconnection.h>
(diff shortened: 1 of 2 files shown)


src/bluetooth/device.hpp
```diff
@@ -99,6 +99,9 @@ class BluetoothDevice: public QObject {
 	Q_PROPERTY(bool batteryAvailable READ batteryAvailable NOTIFY batteryAvailableChanged);
 	/// Battery level of the connected device, from `0.0` to `1.0`. Only valid if @@batteryAvailable is true.
 	Q_PROPERTY(qreal battery READ default NOTIFY batteryChanged BINDABLE bindableBattery);
+	/// Received Signal Strength Indicator in dBm. Only valid when device is discoverable/advertising.
+	/// Typical values range from -30 (very close) to -90 (far away). Values of 0 indicate no signal data.
+	Q_PROPERTY(qint16 rssi READ rssi NOTIFY rssiChanged BINDABLE bindableRssi);
```
outfoxxed: If we're going to expose this at all, it should be normalized onto a linear scale that is easy to make a visual judgement with, and named something simpler, such as "signalStrength".

src/dbus/properties.cpp
```diff
@@ -214,8 +214,10 @@ void DBusPropertyGroup::updatePropertySet(const QVariantMap& properties, bool co
 	}
 }
 
-void DBusPropertyGroup::tryUpdateProperty(DBusPropertyCore* property, const QVariant& variant)
-    const {
+void DBusPropertyGroup::tryUpdateProperty(
+    DBusPropertyCore* property,
+    const QVariant& variant
+) const {
 	property->mExists = true;
```
outfoxxed, line 219: The original formatting is wrong, but leave it because the formatter needs to be reconfigured to fix it.

src/dbus/properties.cpp
```diff
@@ -246,8 +248,17 @@ void DBusPropertyGroup::requestPropertyUpdate(DBusPropertyCore* property) {
...
+			if (!property->isRequired() && error.type() == QDBusError::InvalidArgs
+			    && error.message().contains("No such property"))
+			{
+				qCDebug(logDbusProperties).noquote() << "Error updating optional property" << propStr;
+				qCDebug(logDbusProperties) << error;
+			} else {
+				qCWarning(logDbusProperties).noquote() << "Error updating property" << propStr;
+				qCWarning(logDbusProperties) << error;
+			}
 		} else {
```
outfoxxed, line 256: Please leave this off, it shouldn't be part of an RSSI pr, especially because a direct property update to rssi is never requested.

outfoxxed (changes requested):

src/bluetooth/device.hpp
bbedward: I exposed signalStrength and normalized it 0-100 scale.

src/dbus/properties.cpp
bbedward: undid all changes to this file

src/bluetooth/device.cpp
```diff
@@ -46,12 +47,35 @@ BluetoothDevice::BluetoothDevice(const QString& path, QObject* parent): QObject(
...
+	auto rssiValue = this->bRssi.value();
+	if (rssiValue == 0) {
+		return 0;
+	}
+
+	auto rssi = std::max(static_cast<qint16>(-100), std::min(static_cast<qint16>(-30), rssiValue));
+	auto normalized = static_cast<qint32>(((rssi + 100) / 70.0) * 100.0);
+
+	return std::max(0, std::min(100, normalized));
+}
```
outfoxxed, line 75: dBm is not linear

outfoxxed: Tbh I'm not sure if this should be linear or not. It requires actual testing at various distances to make sure the number is useful

bbedward: I adjusted it, there's no good singular mathematical function I could find to get it as a good representation for display purposes - but I think the numbers are pretty good now.

| dBm | normalized |
|------------|---------------------|
| -30 | 100 |
| -35 | 96 |
| -40 | 92 |
| -45 | 88 |
| -50 | 83 |
| -55 | 79 |
| -60 | 75 |
| -65 | 59 |
| -70 | 42 |
| -75 | 26 |
| -80 | 9 |
| -85 | 5 |
| -90 | 0 |

Pretty much anything > -67 is a good signal, anything -80 or below is usually un-usable. 0 is driver reporting no signal data is available for the device (so treating it as effectively 0 normalized)

purian23:
@outfoxxed can you have a look at the updates made since your last request? Some of these PR's sit for days or weeks at a time with small enhancement requests. I would say many are okay if you co-author as well to help push them along quicker.

outfoxxed:
> @outfoxxed can you have a look at the updates made since your last request? Some of these PR's sit for days or weeks at a time with small enhancement requests. I would say many are okay if you co-author as well to help push them along quicker.

This one is deceptively simple looking. I'm quite busy and have not had the time to test that shown signal strength reports a logical looking value for various devices. The strength calculation function also looks suspiciously like a curve fit better solved by a single equation. Figuring that out for review takes a considerable amount of time which I don't have right now.

bbedward:
Tbh it might be better to just report an enum with Excellent, good, fair, poor or something for the easy to consume value. And raw rssi for those who desire it.

It’s hard to quantify very accurately because different BT versions and protocols and device manufacturers have some variations on the reporting and work better or worse at lower signals/different distances. But I’m good with whatever approach you wanna go with, no strong opinion. 

The main thing for me would be to sort, and show the excellent/good/fair/poor icon/text next to them.

outfoxxed:
I don't have an issue with exposing raw RSSI, and exposing signal strength as an enum/int for general quality is acceptable as well. I would lean towards an integer, e.g. 0-4 for that "quality" value because enums are generally more annoying to deal with in QML.

PR #137, opened 2025-07-22
core/itemimagegrab: `ItemImageGrab` component

soramanew:
Adds a custom QML component `ItemImageGrab` which can be used to grab and save items to an image file. The component provides 2 methods each with 2 overloads, `grab` and `cropAndGrab`.

`grab` is for grabbing and saving an item to a file asynchronously, while `cropAndGrab` does the same but allows cropping to a specific size when saving.

### DIFF
--- a/src/core/instanceinfo.hpp
+++ b/src/core/instanceinfo.hpp
@@ -6,6 +6,14 @@
 #include <sys/types.h>
 
 struct InstanceInfo {
+	Q_GADGET
+
+	Q_PROPERTY(QString instanceId MEMBER instanceId CONSTANT)
+	Q_PROPERTY(QString shellId MEMBER shellId CONSTANT)
+	Q_PROPERTY(QString appId MEMBER appId CONSTANT)
+	Q_PROPERTY(QDateTime launchTime MEMBER launchTime CONSTANT)
+
+public:
 	QString instanceId;
 	QString configPath;
 	QString shellId;
--- a/src/core/qmlglobal.cpp
+++ b/src/core/qmlglobal.cpp
@@ -154,8 +154,8 @@ qint32 QuickshellGlobal::processId() const { // NOLINT
 	return getpid();
 }
 
-QString QuickshellGlobal::instanceId() const { // NOLINT
-	return InstanceInfo::CURRENT.instanceId;
+InstanceInfo QuickshellGlobal::instanceInfo() const { // NOLINT
+	return InstanceInfo::CURRENT;
 }
 
 qsizetype QuickshellGlobal::screensCount(QQmlListProperty<QuickshellScreenInfo>* /*unused*/) {
--- a/src/core/qmlglobal.hpp
+++ b/src/core/qmlglobal.hpp
@@ -17,6 +17,7 @@
 
 #include "../io/processcore.hpp"
 #include "doc.hpp"
+#include "instanceinfo.hpp"
 #include "qmlscreen.hpp"
 
 ///! Accessor for some options under the Quickshell type.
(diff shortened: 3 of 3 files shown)


soramanew:
Unfortunately, I have no idea how to make a custom qml object 😭

I'll try work on this in a few days I guess

src/core/itemimagegrab.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+void ItemImageGrab::grab(const QUrl& path) { this->grab(this->mTarget, path); }
+
+void ItemImageGrab::grab(const QUrl& path, const QSize& targetSize) { this->grab(this->mTarget, path, targetSize); }
+
+void ItemImageGrab::grab(QQuickItem* target, const QUrl& path) { this->grab(target, path, QSize()); }
+
+void ItemImageGrab::grab(QQuickItem* target, const QUrl& path, const QSize& targetSize) {
+    if (!target) return;
```
outfoxxed: if !target log an error

src/core/itemimagegrab.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    QSharedPointer<QQuickItemGrabResult> grabResult;
+    if (targetSize.isEmpty()){
+        grabResult = target->grabToImage();
+    }else{
+        grabResult = target->grabToImage(targetSize);}
+
+    QObject::connect(grabResult.data(), &QQuickItemGrabResult::ready, this, [grabResult, path, this]() {
```
outfoxxed: have you checked if we need to hold the sp, or if its held internally?

src/core/itemimagegrab.hpp
```diff
@@ -0,0 +1,33 @@
...
+    Q_INVOKABLE void grab(QQuickItem* target, const QUrl& path, const QSize& targetSize);
+
+signals:
+    void saved(const QUrl& path);
+    void failed(const QUrl& path);
+
+	void targetChanged();
+
+private:
+	QPointer<QQuickItem> mTarget;
```
outfoxxed: Can't use qpointer, you need to watch the destroy signal to send a targetChanged after nulling

outfoxxed (changes requested):
also format

src/core/itemimagegrab.cpp
soramanew: I have no idea what this means, but it seems to work fine?

soramanew:
> also format

Yeah sorry, forgot about that lol

src/core/itemimagegrab.hpp
```diff
@@ -0,0 +1,25 @@
...
+	Q_INVOKABLE void grab(QQuickItem* target, const QUrl& path);
+	Q_INVOKABLE void grab(QQuickItem* target, const QUrl& path, const QSize& targetSize);
+
+	Q_INVOKABLE void cropAndGrab(QQuickItem* target, const QUrl& path, const QRect& rect);
+	Q_INVOKABLE void
+	cropAndGrab(QQuickItem* target, const QUrl& path, const QRect& rect, const QSize& targetSize);
+
+signals:
+	void saved(const QString& path);
+	void failed(const QUrl& path);
```
HigherOrderLogic: Wouldnt it be better to have both these two be `QString` or `QUrl`?

soramanew: Saved actually gives the local file path so you can use it directly (e.g. doing something to the file after it is saved) while failed gives the original file url you have it so you know which one failed

Hmm saved should probably still give the url as well though

soramanew:
.w.

PR #142, opened 2025-07-23
core/reloader: fix file watcher compatibility with vscode

karboggy:
Some editors (e.g vscode)  can sometimes save files in two steps: truncate + write.
This causes `QFileSystemWatcher` to emit the `fileChanged` signal twice.

It can be visualized with:
`inotifywait -m -e modify,create,delete,delete_self,close_write path/to/my-file.qml`

On the first event (truncate), the file size is zero. If quickshell reloads the configuration at this point, it may result to a `Reloading configuration error`.

This initial event can be safely ignored. The second event (when the file is written) will correctly trigger a reload with valid content.


outfoxxed:
Looks good, but could it reasonably end up writing in multiple steps?

karboggy:
This behavior seems related to all Node.js / Electron apps (including vscode and codium) . For example, VSCode’s save process is discussed in this issue:
https://github.com/microsoft/vscode/issues/63720

src/core/generation.cpp
```diff
@@ -222,6 +222,12 @@ void EngineGeneration::onFileChanged(const QString& name) {
 	if (!this->watcher->files().contains(name)) {
 		this->deletedWatchedFiles.push_back(name);
 	} else {
+		// some editors (e.g vscode) perform file saving in two steps: truncate + write
+		// ignore the first event (truncate) with size 0 to prevent incorrect live reloading
+		if (QFileInfo(name).size() == 0) {
```
outfoxxed: Size is likely also zero if the file does not exist or is a directory. Check .isFile() as well.

outfoxxed (changes requested):

src/core/generation.cpp
karboggy: We can't have a directory  here (in `EngineGeneration::onFileChanged`) as we are connected to the   `QFileSystemWatcher::fileChanged` signal.  Why should  we check `isFile()` ?

outfoxxed: Unlikely but potential race condition. The file not existing is a more important case. Intentional truncation to an empty file is also possible and I'd prefer it be handled but its not that important.

karboggy: I've added the requested `isFile()` check. Is this what you had in mind?

src/core/generation.cpp
```diff
@@ -222,6 +222,13 @@ void EngineGeneration::onFileChanged(const QString& name) {
 	if (!this->watcher->files().contains(name)) {
 		this->deletedWatchedFiles.push_back(name);
 	} else {
+		// some editors (e.g vscode) perform file saving in two steps: truncate + write
+		// ignore the first event (truncate) with size 0 to prevent incorrect live reloading
+		auto fileInfo = QFileInfo(name);
+		if (fileInfo.isFile() && fileInfo.size() == 0) {
+			return;
+		}
```
outfoxxed: Same line return please

outfoxxed (approved):
Otherwise lgtm

src/core/generation.cpp
karboggy: Done

PR #159, opened 2025-07-31
build: add systemd service

axelf4:
Adds a systemd unit configuration file to be installed by CMake.

My motivation is being able to auto-start Quickshell on NixOS/niri with just the following module options:
```nix
systemd.packages = with pkgs; [ quickshell ];
systemd.user.services.quickshell.wantedBy = [ "niri.service" ];
```
A caveat is that the service `PATH` will be limited, meaning `DesktopEntry.execute()` will often not work directly.

C.f.: https://github.com/Alexays/Waybar/blob/0776e694df56c2c849b682369148210d81324e93/resources/waybar.service.in

Tangentially, it would be neat if an `Exec=` key got added to `org.quickshell.desktop`, so it could be placed in `/etc/xdg/autostart`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,3 +157,9 @@ install(
 	DESTINATION ${CMAKE_INSTALL_DATADIR}/icons/hicolor/scalable/apps
 	RENAME org.quickshell.svg
 )
+
+configure_file(assets/quickshell.service.in quickshell.service)
+install(
+	FILES ${CMAKE_BINARY_DIR}/quickshell.service
+	DESTINATION lib/systemd/user
+)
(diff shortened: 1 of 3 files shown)


outfoxxed:
I don't particularly like this because multiple configs are intended to be usable.

axelf4:
Sure, but it could still be valuable as an example or starting point in that case, while being convenient for anyone who only uses their single own config. But if you disagree we will close this.

Nicknamely:
>I don't particularly like this because multiple configs are intended to be usable.
 
 I tried to run quickshell as a service after migrating to UWSM, but I haven't been able to get it working. The config still runs fine from the CLI. See #166  for more details.

axelf4:
Closing due to lack of response.

Issue #147, opened 2025-07-25
Need system tray custom icon support


nx-smul:
<img width="559" height="139" alt="Image" src="https://github.com/user-attachments/assets/e3692723-afdf-4cb7-9975-b92e440cb408" />

custom icons doesn't work in system tray kindly fix it

Kruziikrel13:
This works, haven't had any bad tray icons using this code
```qml
  IconImage {
    source: {
      let icon = root.item.icon;
      if (icon.includes("?path=")) {
        const [name, path] = icon.split("?path=");
        icon = `file://${path}/${name.slice(name.lastIndexOf("/") + 1)}`;
      }
      return icon;
    }
    asynchronous: true
    implicitSize: ShellGlobals.sizes.font.normal * 1.1
  }
```

shadowe1ite:
Any updates ?

PR #185, opened 2025-08-18
core: Icons add searching custom file paths

CMurtagh-LGTM:
Fixes: #147

This is basically just @Kruziikrel13's `QML` code ported to `C++`.
> This works, haven't had any bad tray icons using this code
> 
> ``` QML
>   IconImage {
>     source: {
>       let icon = root.item.icon;
>       if (icon.includes("?path=")) {
>         const [name, path] = icon.split("?path=");
>         icon = `file://${path}/${name.slice(name.lastIndexOf("/") + 1)}`;
>       }
>       return icon;
>     }
>     asynchronous: true
>     implicitSize: ShellGlobals.sizes.font.normal * 1.1
>   }
> ```

The main difference is that Kruziikrel13 does this before qt checks the `QQuickPixmapCache` enabling use of the `QML`'s cache but is overriding the default icon.

### DIFF
--- a/src/core/iconimageprovider.cpp
+++ b/src/core/iconimageprovider.cpp
@@ -19,8 +19,7 @@ IconImageProvider::requestPixmap(const QString& id, QSize* size, const QSize& re
 	if (splitIdx != -1) {
 		iconName = id.sliced(0, splitIdx);
 		path = id.sliced(splitIdx + 6);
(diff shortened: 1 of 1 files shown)


CMurtagh-LGTM:
#219 Has an alternate implementation.

outfoxxed:
thanks

PR #195, opened 2025-08-22
wayland: input method protocol

CMurtagh-LGTM:
- [x] ~~Popup surface~~ This PR is already large, will add in future PR.
- [x] Handle surrounding_text, text_change_cause and content_type events somehow.
- [x] Key repeat
- [x] Documentation

~~Do we want to also expose the virtual keyboard so people can create their own clients?~~

Popup surface and qml virtual keyboard will be in future PRs.

### DIFF
--- a/src/wayland/input_method/keyboard_grab.cpp
+++ b/src/wayland/input_method/keyboard_grab.cpp
@@ -3,6 +3,7 @@
 #include <cstdint>
 #include <vector>
 
+#include <qlogging.h>
 #include <qloggingcategory.h>
 #include <qobject.h>
 #include <qtmetamacros.h>
--- a/src/wayland/input_method/qml.cpp
+++ b/src/wayland/input_method/qml.cpp
@@ -3,7 +3,6 @@
 #include <cstdint>
 #include <utility>
 
-#include <qlogging.h>
 #include <qobject.h>
 #include <qpointer.h>
 #include <qqmlinfo.h>
@@ -30,7 +29,7 @@ void InputMethod::commitString(const QString& text) {
 void InputMethod::setPreeditString() {
 	if (!this->isActive()) return;
 
-	if (mCursorEnd == -1) {
+	if (this->mCursorEnd == -1) {
 		this->handle->setPreeditString(this->mPreeditString, this->mCursorBegin, this->mCursorBegin);
 	} else {
 		this->handle->setPreeditString(this->mPreeditString, this->mCursorBegin, this->mCursorEnd);


outfoxxed:
~~Exposing virtual keyboard would be a good idea, but as a separate PR.~~
Oh I see it was added. Is this ready for review?

CMurtagh-LGTM:
I still need to do a couple things - especially documentation. I'll add the popup screen and exposing virtual keyboard in future prs since this one is already quite large.

CMurtagh-LGTM:
Sorry I haven't updated in a while.
*This* pull request is good to go.

I have a separate branch that I have been working on the adding the popup surface functionality however it looks like I need to use the private qt headers, which I have not yet been able to work out.

outfoxxed:
Looks like CI is broken here, can you rebase on master? For private qt headers you should just be able to import them if you add a dependency on the private qt module.

src/wayland/input_method/input-method-unstable-v2.xml
outfoxxed, line 1: Why is this protocol included directly instead of from w-p?

src/wayland/input_method/qml.hpp
```diff
@@ -0,0 +1,160 @@
...
+	Q_PROPERTY(QMLContentHint::Enum contentHint READ contentHint NOTIFY contentHintChanged);
+	/// The content_purpose of the text input see https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:enum:content_purpose
+	Q_PROPERTY(QMLContentPurpose::Enum contentPurpose READ contentPurpose NOTIFY contentPurposeChanged);
+	/// Clear the preedit text when the keyboard grab is released
+	Q_PROPERTY(
+	    bool clearPreeditOnKeyboardRelease MEMBER mClearPreeditOnKeyboardRelease NOTIFY
+	        clearPreeditOnKeyboardReleaseChanged
+	);
+	/// The @@Keyboard$ that will handle the grabbed keyboard.
+	/// Use @@KeyboardTextEdit$ for most cases.
```
outfoxxed: You can drop the $ if there's a space, period, or comma after the reference

src/wayland/input_method/qml.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+	/// Sends a string to the text input currently focused
+	Q_INVOKABLE void sendString(const QString& text);
+	/// Provides virtual text in the text input so the user can visualise what they write
+	/// This will override any previous preedit text
+	/// If `cursorBegin == cursorEnd == -1` the text input will not show a cursor
+	/// If `cursorBegin == cursorEnd == n` the text input will show a cursor at n
+	/// If `cursorBegin == n` and `cursorEnd == m` the text from n to m will be highlighted
+	Q_INVOKABLE void
+	sendPreeditString(const QString& text, int32_t cursorBegin = -1, int32_t cursorEnd = -1);
```
outfoxxed: The documentation generator doesn't properly support default args, so probably make two forms of this: `sendPreeditString(text, begin, end)`, `sendPreeditString(text)`

src/wayland/input_method/qml.hpp
```diff
@@ -0,0 +1,160 @@
...
+	/// @@hasInput$
+	Q_INVOKABLE [[nodiscard]] bool hasInput() const;
+	/// Retries getting the input method.
+	Q_INVOKABLE void getInput();
+	/// Releases the input method so another program can use it.
+	Q_INVOKABLE void releaseInput();
+
+	/// @@hasKeyboard$
+	Q_INVOKABLE [[nodiscard]] bool hasKeyboard() const;
+	/// Grabs the current keyboard so the input can be intercepted by the @@keyboard$ object
```
outfoxxed: Property getters dont need to be invokable

src/wayland/input_method/qml.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+	/// @@hasInput$
+	Q_INVOKABLE [[nodiscard]] bool hasInput() const;
+	/// Retries getting the input method.
+	Q_INVOKABLE void getInput();
+	/// Releases the input method so another program can use it.
+	Q_INVOKABLE void releaseInput();
+
+	/// @@hasKeyboard$
+	Q_INVOKABLE [[nodiscard]] bool hasKeyboard() const;
```
outfoxxed: Property getters dont need to be invokable

src/wayland/input_method/keyboard_grab.cpp
```diff
@@ -0,0 +1,195 @@
...
+    uint32_t /*time*/,
+    uint32_t key,
+    uint32_t state
+) {
+	if (serial <= this->mSerial) return;
+	this->mSerial = serial;
+
+	key += WAYLAND_KEY_OFFSET;
+
+#if INPUT_METHOD_PRINT
```
outfoxxed: I am glad you added logs but instead of ifdefing them, use logging categories (grep for QS_LOGGING_CATEGORY and qCDebug)

src/wayland/input_method/keyboard_grab.cpp
```diff
@@ -0,0 +1,195 @@
...
+	);
+
+	// Tell the text input to release all the keys
+	for (xkb_keycode_t key = 0; key < this->mKeyState.size(); ++key) {
+		this->mVirturalKeyboard->sendKey(
+		    key + this->mKeyMapState.minKeycode(),
+		    WL_KEYBOARD_KEY_STATE_RELEASED
+		);
+	}
+}
```
outfoxxed, line 91: Why don't we use QXkbCommon for keysym handling similarly to how its handled for wl_keyboard in qtbase? See `qwaylandinputdevice.cpp` Holding off on reviewing keymap related things for now.

src/wayland/input_method/input_method.cpp
```diff
@@ -0,0 +1,124 @@
...
+) {
+	this->set_preedit_string(text, cursorBegin, cursorEnd);
+}
+void InputMethodHandle::deleteText(int before, int after) {
+	this->zwp_input_method_v2::delete_surrounding_text(before, after);
+}
+void InputMethodHandle::commit() { this->zwp_input_method_v2::commit(this->serial++); }
+
+bool InputMethodHandle::hasKeyboard() const { return this->keyboard != nullptr; }
+QPointer<InputMethodKeyboardGrab> InputMethodHandle::grabKeyboard() {
```
outfoxxed: There is no point in returning a QPointer. QPointers are registered to the QObject on creation to receive a hook on its destruction and null themselves. As the hook can't do anything else they're generally not very useful, but they are never useful at all in this position.

src/wayland/input_method/input_method.cpp
```diff
@@ -0,0 +1,124 @@
...
+	if (this->keyboard) return this->keyboard;
+
+	this->keyboard = new InputMethodKeyboardGrab(this, this->grab_keyboard());
+
+	return this->keyboard;
+}
+void InputMethodHandle::releaseKeyboard() {
+	if (!this->keyboard) return;
+	this->keyboard->deleteLater();
```
outfoxxed: Should just be able to use delete.

src/wayland/input_method/input_method.cpp
```diff
@@ -0,0 +1,124 @@
...
+	}
+}
+
+void InputMethodHandle::zwp_input_method_v2_unavailable() {
+	if (!this->mAvailable) return;
+	this->mAvailable = false;
+	InputMethodManager::instance()->releaseInput();
+	qDebug()
+	    << "Compositor denied input method request, likely due to one already existing elsewhere";
+}
```
outfoxxed, line 121: This case probably isn't this uncommon if the user actually has an IME, and should be reported back in a useful way if it isn't already.

src/wayland/input_method/manager.cpp
```diff
@@ -0,0 +1,59 @@
...
+QPointer<InputMethodHandle> InputMethodManager::acquireInput() {
+	if (this->inputMethod && this->inputMethod->isAvailable()) return this->inputMethod;
+
+	this->inputMethod = new InputMethodHandle(this, this->get_input_method(getSeat()));
+
+	return this->inputMethod;
+}
+
+void InputMethodManager::releaseInput() {
+	this->inputMethod->deleteLater();
```
outfoxxed: Should not require more than delete.

src/wayland/input_method/manager.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+InputMethodManager::InputMethodManager(): QWaylandClientExtensionTemplate(1) { this->initialize(); }
+
+InputMethodManager::~InputMethodManager() { this->destroy(); }
+
+InputMethodManager* InputMethodManager::instance() {
+	// The OS should free this memory when we exit
+	static auto* instance = new InputMethodManager();
+	return instance;
+}
```
outfoxxed, line 25: Comment is unnnecessary.

Return nullptr when not initialized.

src/wayland/input_method/qml.cpp
```diff
@@ -0,0 +1,206 @@
...
+	this->handle->sendPreeditString(text, cursorBegin, cursorEnd);
+	this->handle->commit();
+}
+
+void InputMethod::deleteText(int before, int after) {
+	if (!this->isActive()) return;
+
+	this->handle->deleteText(before, after);
+	this->handle->commit();
+}
```
outfoxxed: I would prefer commit to be separated for this one. The pre-edit string and cursor region may be better exposed as properties instead of a function as well.

src/wayland/input_method/qml.cpp
```diff
@@ -0,0 +1,206 @@
...
+	this->mKeyboardComponent = keyboardComponent;
+	emit this->keyboardComponentChanged();
+
+	if (this->keyboard) {
+		this->keyboard->deleteLater();
+		this->keyboard = nullptr;
+	}
+
+	this->handleKeyboardActive();
+}
```
outfoxxed: Why are components even involved in this? Shouldn't we just expose the keyboard as a member of the inputmethod which is null if we haven't grabbed it?

src/wayland/input_method/types.cpp
```diff
@@ -0,0 +1,125 @@
...
+	if (contentHint & Multiline) {
+		if (!first) {
+			string += " | ";
+		}
+		string += "Multiline";
+		first = false;
+	}
+
+	if (string == "") string = "None";
+	return string;
```
outfoxxed, line 57: Make a list and join it instead of if (!first)?

src/wayland/input_method/virtual-keyboard-unstable-v1.xml
outfoxxed, line 1: 1. Might be better to split this
2. Include from w-p instead?

src/wayland/input_method/input-method-unstable-v2.xml
CMurtagh-LGTM: It is kind of confusing as there are three versions of input-method floating around.
- zwp_input_method_v1 - unstable
- xx_input_method_v1 - experimental
- zwp_input_method_v2 - external
Hyprland uses zwp_input_method_v2 so I went with that one.
zwp_input_method_v2 is not included in Wayland protocols.

src/wayland/input_method/input_method.cpp
CMurtagh-LGTM: I don't know of a better way to report it than a log message.

src/wayland/input_method/virtual-keyboard-unstable-v1.xml
CMurtagh-LGTM: 1. I'm unsure what you mean by "split this".
2. This protocol is not inside Wayland protocols unfortunately.

src/wayland/input_method/keyboard_grab.cpp
CMurtagh-LGTM: I had a look at the private header QXkbCommon, it doesn't have any functions that would be useful here.

If you think there is a particular function I should be using then please tell me.

src/wayland/input_method/qml.cpp
CMurtagh-LGTM: I don't understand why you want commit to be separate for only delete.

CMurtagh-LGTM (commented):
I still need to have a look at adding qml test files.

src/wayland/input_method/virtual-keyboard-unstable-v1.xml
outfoxxed: 1. It *seems* pretty independent from input method and looks like it makes sense to pr separately and not in the input method dir
2. Unfortunate, fine to vendor then

src/wayland/input_method/input-method-unstable-v2.xml
outfoxxed: I'd gotten it confused with xx-input since they seem pretty similar in most respects. Diffed them and it looks like xx can't grab the keyboard properly and zwp v2 can't position its surface properly. Beautiful.

Fine to vendor it then I guess

src/wayland/input_method/keyboard_grab.cpp
outfoxxed: Ok so I've gone and looked through this and QWaylandInputDevice again.

I am confused why this involves virtual keyboard at all as it seems like it should just be input. The protocol says its an exclusive grab so we shouldnt be passing input anywhere else either. Many compositors have only one seat in which case that will do nothing at all.

What I was thinking for the keymap is that we could basically mirror the wl_keyboard implementation in qwaylandinputdevice which uses qxkbcommon to derive strings, instead of duplicating that logic, But now I'm not sure what this is intended to do.

outfoxxed:
Responded inline to the comments above but github seems to have eaten them. They're in the files still

src/wayland/input_method/keyboard_grab.cpp
CMurtagh-LGTM: https://dorotac.eu/posts/input_method/ explains what is going on with the virtual keyboard.

CMurtagh-LGTM:
This protocol is likely to change in the future, I'd prefer to wait for things to settle down before merging something in.
See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/405.

PR #206, opened 2025-08-29
core/colorquant: add `imageRect` option for cropping image

HigherOrderLogic:
Allow users to crop the image and run the color quantizer on just a part of the source image.


kossLAN:
I see no issues with this, so it has my support.

PR #246, opened 2025-09-18
src/desktopentry: fix handling of hidden entries

bbedward:
Hidden desktop entries currently get parsed out, however all lower-ranked entires remain. So for example if I:
- create `~/.local/share/applications/org.mozilla.Thunderbird.desktop` with `Hidden=true`, it gets ignored by the parser. However the entry in `/usr/share/applications/org.mozilla.Thunderbird.desktop` does not get ignored.
- Expectation would be that all instances of thunderbird get hidden.

This fixes it by:
- Carrying `hidden` through the parsing stage
- Removing all hidden entries in the scanner when it comes across on.


src/core/desktopentry.cpp
```diff
@@ -178,11 +180,21 @@ ParsedDesktopEntryData DesktopEntry::parseText(const QString& id, const QString&
...
+				// Prefer unlocalized (invalid locale) when both scores tie (usually 0)
+				if (!locale.isValid() && old.first.isValid()) {
+					shouldReplace = true;
+				}
+			}
 
-			if (newScore > oldScore || (oldScore == 0 && !locale.isValid())) {
+			if (shouldReplace) {
 				entries.insert(key, qMakePair(locale, value));
 			}
```
outfoxxed, line 189: What has meaningfully changed here? Returning zero in matchScore should always force fallback.

src/core/desktopentry.cpp
```diff
@@ -495,6 +507,21 @@ void DesktopEntryManager::onScanCompleted(const QList<ParsedDesktopEntryData>& s
...
+			newLowercaseEntries.remove(lowerId);
+
+			if (auto it = oldEntries.find(data.id); it != oldEntries.end()) {
+				it.value()->deleteLater();
+				oldEntries.erase(it);
+			}
+
+			qCDebug(logDesktopEntry) << "Masking hidden desktop entry" << data.id;
+			continue;
+		}
```
outfoxxed, line 513: 1. Why deleteLater?
2. Is hidden supposed to match only things lower than it in the path, or things higher than it as well? Because this looks like its only matching things lower than it.

src/core/desktopentry.cpp
bbedward: I reverted this part, maybe not much just some explicitness - but its not relevant to the core bug fix

src/core/desktopentry.cpp
bbedward: 1. I just thought it's safer, if the object has any signals connected to ir and what not.
2. Lower than it is the intention, that is my understanding of the specification

I mean my understanding of it is that only the highest priority desktop entry should be considered - so it shouldnt treat higher-prio entries as deleted.

PR #277, opened 2025-09-29
wayland/shortcuts-inhibit: add shortcuts inhibitor

bbedward:
Implement [keyboard-shortcuts-inhibit-unstable-v1](https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1)

There's probably a few use cases for this in quickshell, but the one i'm interested in is getting an interface for creating keybinds. Without it it gets eaten by the compositor.

Usage is as follows:

```
import Quickshell.Wayland

ShortcutsInhibitor {
    window: root // PanelWindow
    enabled: root.recording
}
```

### DIFF
--- a/src/wayland/shortcuts_inhibit/inhibitor.cpp
+++ b/src/wayland/shortcuts_inhibit/inhibitor.cpp
@@ -1,0 +1,187 @@
+#include "inhibitor.hpp"
+
+#include <private/qwaylandwindow_p.h>
+#include <qguiapplication.h>
+#include <qlogging.h>
+#include <qloggingcategory.h>
+#include <qobject.h>
+#include <qtmetamacros.h>
+#include <qwindow.h>
+
+#include "../../window/proxywindow.hpp"
+#include "../../window/windowinterface.hpp"
+#include "proto.hpp"
+
+namespace qs::wayland::shortcuts_inhibit {
+using QtWaylandClient::QWaylandWindow;
+
+ShortcutInhibitor::ShortcutInhibitor() {
+	this->bBoundWindow.setBinding([this] {
+		return this->bEnabled ? this->bWindowObject.value() : nullptr;
+	});
+
+	this->bActive.setBinding([this]() {
+		auto* inhibitor = this->bInhibitor.value();
+		if (!inhibitor) return false;
+		if (!inhibitor->bindableActive().value()) return false;
+		return this->bWindow.value() == this->bFocusedWindow;
+	});
+
+	QObject::connect(
+	    dynamic_cast<QGuiApplication*>(QGuiApplication::instance()),
+	    &QGuiApplication::focusWindowChanged,
+	    this,
+	    &ShortcutInhibitor::onFocusedWindowChanged
+	);
+
+	this->onFocusedWindowChanged(QGuiApplication::focusWindow());
+}
+
+ShortcutInhibitor::~ShortcutInhibitor() {
+	if (!this->bInhibitor) return;
+
+	auto* manager = impl::ShortcutsInhibitManager::instance();
+	if (!manager) return;
+
+	manager->unrefShortcutsInhibitor(this->bInhibitor);
+}
+
+void ShortcutInhibitor::onBoundWindowChanged() {
+	auto* window = this->bBoundWindow.value();
+	auto* proxyWindow = qobject_cast<ProxyWindowBase*>(window);
+
+	if (!proxyWindow) {
+		if (auto* iface = qobject_cast<WindowInterface*>(window)) {
+			proxyWindow = iface->proxyWindow();
+		}
+	}
+
+	if (proxyWindow == this->proxyWindow) return;
+
+	if (this->proxyWindow) {
+		QObject::disconnect(this->proxyWindow, nullptr, this, nullptr);
+		this->proxyWindow = nullptr;
+	}
+
+	if (this->mWaylandWindow) {
+		QObject::disconnect(this->mWaylandWindow, nullptr, this, nullptr);
+		this->mWaylandWindow = nullptr;
+		this->onWaylandSurfaceDestroyed();
+	}
+
+	if (proxyWindow) {
+		this->proxyWindow = proxyWindow;
+
+		QObject::connect(proxyWindow, &QObject::destroyed, this, &ShortcutInhibitor::onWindowDestroyed);
+
+		QObject::connect(
+		    proxyWindow,
+		    &ProxyWindowBase::backerVisibilityChanged,
+		    this,
+		    &ShortcutInhibitor::onWindowVisibilityChanged
+		);
+
+		this->onWindowVisibilityChanged();
+	}
+}
+
+void ShortcutInhibitor::onWindowDestroyed() {
+	this->proxyWindow = nullptr;
+	this->onWaylandSurfaceDestroyed();
+}
+
+void ShortcutInhibitor::onWindowVisibilityChanged() {
+	if (!this->proxyWindow->isVisibleDirect()) return;
+
+	auto* window = this->proxyWindow->backingWindow();
+	if (!window->handle()) window->create();
+
+	auto* waylandWindow = dynamic_cast<QWaylandWindow*>(window->handle());
+	if (!waylandWindow) {
+		qCCritical(impl::logShortcutsInhibit()) << "Window handle is not a QWaylandWindow";
+		return;
+	}
+	if (waylandWindow == this->mWaylandWindow) return;
+	this->mWaylandWindow = waylandWindow;
+	this->bWindow = window;
+
+	QObject::connect(
+	    waylandWindow,
+	    &QObject::destroyed,
+	    this,
+	    &ShortcutInhibitor::onWaylandWindowDestroyed
+	);
+
+	QObject::connect(
+	    waylandWindow,
+	    &QWaylandWindow::surfaceCreated,
+	    this,
+	    &ShortcutInhibitor::onWaylandSurfaceCreated
+	);
+
+	QObject::connect(
+	    waylandWindow,
+	    &QWaylandWindow::surfaceDestroyed,
+	    this,
+	    &ShortcutInhibitor::onWaylandSurfaceDestroyed
+	);
+
+	if (waylandWindow->surface()) this->onWaylandSurfaceCreated();
+}
+
+void ShortcutInhibitor::onWaylandWindowDestroyed() { this->mWaylandWindow = nullptr; }
+
+void ShortcutInhibitor::onWaylandSurfaceCreated() {
+	auto* manager = impl::ShortcutsInhibitManager::instance();
+
+	if (!manager) {
+		qWarning() << "Cannot enable shortcuts inhibitor as keyboard-shortcuts-inhibit-unstable-v1 is "
+		              "not supported by "
+		              "the current compositor.";
+		return;
+	}
+
+	if (this->bInhibitor) {
+		qFatal("ShortcutsInhibitor: inhibitor already exists when creating surface");
+	}
+
+	this->bInhibitor = manager->createShortcutsInhibitor(this->mWaylandWindow);
+}
+
+void ShortcutInhibitor::onWaylandSurfaceDestroyed() {
+	if (!this->bInhibitor) return;
+
+	auto* manager = impl::ShortcutsInhibitManager::instance();
+	if (!manager) return;
+
+	manager->unrefShortcutsInhibitor(this->bInhibitor);
+	this->bInhibitor = nullptr;
+}
+
+void ShortcutInhibitor::onInhibitorChanged() {
+	auto* inhibitor = this->bInhibitor.value();
+	if (inhibitor) {
+		QObject::connect(
+		    inhibitor,
+		    &impl::ShortcutsInhibitor::activeChanged,
+		    this,
+		    &ShortcutInhibitor::onInhibitorActiveChanged
+		);
+	}
+}
+
+void ShortcutInhibitor::onInhibitorActiveChanged() {
+	auto* inhibitor = this->bInhibitor.value();
+	if (inhibitor && !inhibitor->isActive()) {
+		// Compositor has deactivated the inhibitor, making it invalid.
+		// Set enabled to false so the user can enable it again to create a new inhibitor.
+		this->bEnabled = false;
+		emit this->cancelled();
+	}
+}
+
+void ShortcutInhibitor::onFocusedWindowChanged(QWindow* focusedWindow) {
+	this->bFocusedWindow = focusedWindow;
+}
+
+} // namespace qs::wayland::shortcuts_inhibit
--- a/src/wayland/shortcuts_inhibit/inhibitor.hpp
+++ b/src/wayland/shortcuts_inhibit/inhibitor.hpp
@@ -1,0 +1,89 @@
+#pragma once
+
+#include <qobject.h>
+#include <qproperty.h>
+#include <qqmlintegration.h>
+#include <qtclasshelpermacros.h>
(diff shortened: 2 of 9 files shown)


outfoxxed:
This does not look helpful for creating keybinds and I can't think of much to use it for apart from making misguided lock screens that don't work properly.

Can you explain exactly what you want to do with it and why having your compositor run qs ipc does not solve it?

bbedward:
> This does not look helpful for creating keybinds and I can't think of much to use it for apart from making misguided lock screens that don't work properly.
> 
> Can you explain exactly what you want to do with it and why having your compositor run qs ipc does not solve it?

Specifically, I want to integrate a settings screen for configuring niri keybinds into my shell. Which is now possible with niris config include branch. 

It's not possible to have a keybind input field without it. (If I want to overwrite one that's already bound)

bbedward:
https://github.com/user-attachments/assets/27807fb3-d76b-4176-b9ef-5ad55189ba6d

This is the type of input field I have in mind

bbedward:
I added an `activeChanged` signal and also accounted for creating multiple on the same surface

```
WARN quickshell.wayland.shortcuts_inhibit: An inhibitor already exists for surface, skipping creation. 0x7fe132c6bc40
```

src/wayland/shortcuts_inhibit/inhibitor.hpp
```diff
@@ -32,6 +32,16 @@ class ShortcutsInhibitor: public QObject {
...
 	Q_PROPERTY(QObject* window READ window WRITE setWindow NOTIFY windowChanged);
+	/// Whether the inhibitor is currently active. The inhibitor is active when the compositor
+	/// has granted the request and shortcuts are being inhibited.
+	///
+	/// The compositor may deactivate the inhibitor if the user requests normal shortcuts to be restored.
+	/// When inactive, there is no way to programmatically reactivate it - the user must do so through
+	/// compositor-specific mechanisms.
+	///
+	/// Note that if the surface loses focus or is destroyed, the inhibitor becomes irrelevant and this
+	/// property will become false, but no inactive event is sent by the compositor in those cases.
```
outfoxxed: Users do not need to know about the wayland events, just that its only valid if the window has keyboard focus.

src/wayland/shortcuts_inhibit/proto.cpp
```diff
@@ -36,15 +37,42 @@ ShortcutsInhibitManager::createShortcutsInhibitor(QtWaylandClient::QWaylandWindo
...
 	}
 
+	auto* wlSurface = surface->surface();
+
+	if (this->inhibitors.contains(wlSurface)) {
+		qCWarning(logShortcutsInhibit)
+		    << "An inhibitor already exists for surface, skipping creation." << wlSurface;
+		return nullptr;
+	}
+
```
outfoxxed, line 47: Multiple inhibitors targeting a given surface should be merged, not prevented.

bbedward: Added merging

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+	});
+}
+
+ShortcutsInhibitor::~ShortcutsInhibitor() {
+	if (!this->inhibitor) return;
+
+	auto* manager = impl::ShortcutsInhibitManager::instance();
+	if (!manager) return;
+
+	QObject::disconnect(this->inhibitor, nullptr, this, nullptr);
```
outfoxxed: QObject destruction does this automatically

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+void ShortcutsInhibitor::boundWindowChanged() {
+	auto* window = this->bBoundWindow.value();
+	auto* proxyWindow = qobject_cast<ProxyWindowBase*>(window);
+
+	if (proxyWindow == nullptr) {
+		if (auto* iface = qobject_cast<WindowInterface*>(window)) {
+			proxyWindow = iface->proxyWindow();
+		}
+	}
```
outfoxxed: If we set bWindowObject, and this triggers boundWindowChanged, proxyWindow should not be set in setWindow but instead only the binding. Additionally, bindable properties internally diff state so you dont need the == check if you're just setting a bindable

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+	this->bWindowObject = nullptr;
+}
+
+void ShortcutsInhibitor::onWindowVisibilityChanged() {
+	if (!this->proxyWindow->isVisibleDirect()) return;
+
+	auto* window = this->proxyWindow->backingWindow();
+	if (!window->handle()) window->create();
+
+	auto* waylandWindow = dynamic_cast<QWaylandWindow*>(window->handle());
```
outfoxxed, line 99: This can be null if not running on wayland

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+		              "not supported by "
+		              "the current compositor.";
+		return;
+	}
+
+	if (this->inhibitor) {
+		QObject::disconnect(this->inhibitor, nullptr, this, nullptr);
+		manager->unrefShortcutsInhibitor(this->inhibitor);
+		this->inhibitor = nullptr;
+	}
```
outfoxxed: Inhibitor should never be initialized here. If anything this should be a qFatal. Its lifetime should end with the wayland surface.

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+		    proxyWindow,
+		    &ProxyWindowBase::backerVisibilityChanged,
+		    this,
+		    &ShortcutsInhibitor::onWindowVisibilityChanged
+		);
+
+		this->onWindowVisibilityChanged();
+	}
+
+	emit this->windowChanged();
```
outfoxxed: It looks like `window` and `bBoundWindow` should be pretty much unrelated. As window hasn't changed here, it shouldn't update. If the window is actually connected should be transparent to the `window` property.

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+	this->inhibitor = manager->createShortcutsInhibitor(this->mWaylandWindow);
+
+	if (this->inhibitor) {
+		QObject::connect(
+		    this->inhibitor,
+		    &impl::ShortcutsInhibitor::activeChanged,
+		    this,
+		    &ShortcutsInhibitor::activeChanged
+		);
+	}
```
outfoxxed: If the inhibitor already existed and we're just +1ing a ref, it might already be active, and we should update the active prop if so.

The best way to handle all the cases here is to make active a bindable property on the proto inhibitor, then bind that to another bindable on qml object.

src/wayland/shortcuts_inhibit/proto.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+	ShortcutsInhibitor* createShortcutsInhibitor(QtWaylandClient::QWaylandWindow* surface);
+	void refShortcutsInhibitor(ShortcutsInhibitor* inhibitor);
+	void unrefShortcutsInhibitor(ShortcutsInhibitor* inhibitor);
+
+	static ShortcutsInhibitManager* instance();
+
+private:
+	QHash<wl_surface*, ShortcutsInhibitor*> inhibitors;
+	QHash<ShortcutsInhibitor*, int> refCounts;
```
outfoxxed: Why not `QHash<wl_surface*, QPair<ShortcutsInhibitor*, int>>`?

src/wayland/shortcuts_inhibit/proto.cpp
```diff
@@ -0,0 +1,98 @@
...
+		return inhibitor;
+	}
+
+	auto* inhibitor =
+	    new ShortcutsInhibitor(this->inhibit_shortcuts(wlSurface, inputDevice->object()), wlSurface);
+	this->inhibitors.insert(wlSurface, inhibitor);
+	this->refCounts.insert(inhibitor, 1);
+	qCDebug(logShortcutsInhibit) << "Created inhibitor" << inhibitor << "for surface" << wlSurface;
+	return inhibitor;
+}
```
outfoxxed, line 53: We should return an existing inhibitor if we have one that matches.

src/wayland/shortcuts_inhibit/inhibitor.cpp
bbedward: removed

src/wayland/shortcuts_inhibit/inhibitor.cpp
bbedward: simplified to only set binding

src/wayland/shortcuts_inhibit/inhibitor.cpp
bbedward: Added a null check

src/wayland/shortcuts_inhibit/inhibitor.cpp
bbedward: Made active a bindable

src/wayland/shortcuts_inhibit/proto.hpp
bbedward: Refactored to use QPair

src/wayland/shortcuts_inhibit/proto.cpp
bbedward: It is if it already exists for the surface

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -147,19 +128,15 @@ void ShortcutsInhibitor::onWaylandSurfaceCreated() {
...
 	if (this->inhibitor) {
-		QObject::connect(
-		    this->inhibitor,
-		    &impl::ShortcutsInhibitor::activeChanged,
-		    this,
-		    &ShortcutsInhibitor::activeChanged
-		);
+		this->bActive.setBinding([this]() {
+			return this->inhibitor ? this->inhibitor->bindableActive().value() : false;
+		});
```
outfoxxed: inhibitor must be a bindable for this to work, and there shouldnt be a reason to set this anywhere but the constructor.

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -107,6 +87,7 @@ void ShortcutsInhibitor::onWindowVisibilityChanged() {
 	if (!window->handle()) window->create();
 
 	auto* waylandWindow = dynamic_cast<QWaylandWindow*>(window->handle());
+	if (!waylandWindow) return;
```
outfoxxed: add a qCCritical

src/wayland/shortcuts_inhibit/inhibitor.hpp
```diff
@@ -50,7 +50,7 @@ class ShortcutsInhibitor: public QObject {
 	void setWindow(QObject* window);
 
 	[[nodiscard]] QBindable<bool> bindableEnabled() { return &this->bEnabled; }
-	[[nodiscard]] bool isActive() const;
+	[[nodiscard]] QBindable<bool> bindableActive() { return &this->bActive; }
```
outfoxxed: readonly bindables should be const fns, which removes the bindable's setter.

src/wayland/shortcuts_inhibit/proto.cpp
```diff
@@ -40,40 +40,46 @@ ShortcutsInhibitManager::createShortcutsInhibitor(QtWaylandClient::QWaylandWindo
...
+	if (!inhibitor) return;
+
+	auto* surface = inhibitor->surface();
+	if (!this->inhibitors.contains(surface)) return;
+
+	auto& pair = this->inhibitors[surface];
+	pair.second++;
+	qCDebug(logShortcutsInhibit) << "Incremented refcount for inhibitor" << inhibitor
+	                             << "- refcount:" << pair.second;
 }
```
outfoxxed: This is still unused as far as a I see

outfoxxed (changes requested):

src/wayland/shortcuts_inhibit/inhibitor.cpp
bbedward: Made it bindable & moved to constructor

src/wayland/shortcuts_inhibit/proto.cpp
bbedward: removed

src/wayland/shortcuts_inhibit/proto.hpp
```diff
@@ -37,21 +39,24 @@ class ShortcutsInhibitor
...
 	[[nodiscard]] wl_surface* surface() const { return this->mSurface; }
 
 signals:
 	void activeChanged();
 
 protected:
+	// NOLINTBEGIN(readability-identifier-naming)
 	void zwp_keyboard_shortcuts_inhibitor_v1_active() override;
 	void zwp_keyboard_shortcuts_inhibitor_v1_inactive() override;
+	// NOLINTEND(readability-identifier-naming)
```
outfoxxed: This lint shouldnt trigger in the first place for overrides

bbedward: Removed, I might not have tidyfox configured right - lint-changed reports some errors for me that the CI doesn't.

outfoxxed: All that lint does is thisptr usage. Depending on your version of clang-tidy vs ci's you might get different reports.

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+namespace qs::wayland::shortcuts_inhibit {
+using QtWaylandClient::QWaylandWindow;
+
+ShortcutsInhibitor::ShortcutsInhibitor()
+    : bEnabled(false)
+    , bWindowObject(nullptr)
+    , bBoundWindow(nullptr)
+    , bInhibitor(nullptr)
+    , bActive(false) {
```
outfoxxed: These are all redundant

src/wayland/shortcuts_inhibit/inhibitor.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+	auto* manager = impl::ShortcutsInhibitManager::instance();
+	if (!manager) return;
+
+	manager->unrefShortcutsInhibitor(this->bInhibitor);
+}
+
+QObject* ShortcutsInhibitor::window() const { return this->bWindowObject; }
+
+void ShortcutsInhibitor::setWindow(QObject* window) { this->bWindowObject = window; }
```
outfoxxed: This can just be a BINDABLE declaration in the header

outfoxxed (commented):
Don't see anything else wrong. This is missing a module.md file though, which is what allows the documentation generator to find the module.

outfoxxed (approved):
~~Tested under hyprland, sway, and niri. Behavior seems acceptable.~~

Edit: Almost, theres still a few cases that are wrong and don't match the documentation:

1. Active is not dependent on the window having keyboard focus
2. Enabled is not set to false if the compositor sends inactive() (an additional inactive() signal would also be helpful here)

outfoxxed:
I've fixed active. Will merge after CI passes.

outfoxxed:
Thanks

PR #302, opened 2025-10-11
Add service module to write PolKit agents

### LINKED ISSUE #271: [FEATURE]: Polkit Authentication Agent
In summary, [Polkit](https://wiki.archlinux.org/title/Polkit) is an application-level toolkit to grant privileged access to graphical applications.

An **Authentication Agent** is a graphical prompt that requires your password to grant privileges (like `sudo`) to graphical applications.

[polkit-kde-agent](https://archlinux.org/packages/extra/x86_64/polkit-kde-agent/) example:

Image

---

Would be nice if **Quickshell** had an **API** to create an **authentication agent** like this.

Cu3PO42:
This implements a new service module that enables users to write PolKit agents by integrating with `libpolkit` and `libpolkitagent`. A test configuration is added to the source.

**Status**: I've done everything I have planned.

<details>
<summary>Original description</summary>

This implements a new service module that enables users to write PolKit agents. To this end, I'm integrating with the existing `polkit-qt-1` library that is also used by the KDE agent.

A relatively minimal example configuration using this new service can be found [here](https://gist.github.com/Cu3PO42/12675e61f9e497d22313e12c09117852).

There's a minor TODO left around cleanup of the `Identity` objects that I haven't been able to get quite right due to my inexperience with Qt. I'm hoping you might have a clean solution for this.

</details>

Closes #271.


outfoxxed:
While I am not opposed to the addition of a polkit integration, I would rather avoid all KDE dependencies as including them leads to massive dependency trees on most distros, and they're usually pretty replaceable.

polkit-qt-1 appears to be a very thin wrapper around the polkit api that doesn't provide much. I would prefer to avoid it unless you have a good reason to pull it in.

Cu3PO42:
polkit-qt-1 is indeed a very thin wrapper. The 'good' reason for using it is that it deals with some of the more error-prone aspects of the upstream libraries and doesn't really bring anything that we don't need. While it is maintained by the KDE team, the package doesn't actually pull in anything but Polkit itself and Qt on the distros I checked (Debian, Ubuntu, Arch, Fedora, RHEL, NixOS, or Gentoo). I find it unlikely that others would have wildly different dependencies.

That said, it should also be fairly easy to not use it at the cost of a few hundred LOC, which would have the benefit of me getting to correct one of the limitations of polkit-qt-1.

outfoxxed:
It sounds like one of the more reasonable ones if we had to use it then, though I would personally still prefer not to.

Cu3PO42:
Alright. I'll see what I can do.

Cu3PO42:
I removed the dependency on polkit-qt, interacting with the PolKit libraries directly, instead.

src/services/polkit/qml.hpp
```diff
@@ -0,0 +1,286 @@
...
+	explicit InputRequest(QString message, bool echo, QObject* parent = nullptr);
+	~InputRequest() override;
+
+	[[nodiscard]] const QString& message() const;
+	[[nodiscard]] bool echo() const;
+
+private:
+	QString mMessage;
+	bool mEcho;
+};
```
outfoxxed: Why not use Q_GADGETs?

src/services/polkit/qml.hpp
```diff
@@ -0,0 +1,286 @@
...
+	Q_OBJECT;
+	QML_ELEMENT;
+	Q_INTERFACES(QQmlParserStatus);
+
+	// clang-format off
+    /// The D-Bus path that this agent listener will use.
+    ///
+    /// > [!INFO] This value can be set to any valid path and has no impact on
+    /// > the functionality of the listener.
+    Q_PROPERTY(QString path READ path WRITE setPath REQUIRED);
```
outfoxxed: 1. We should have a default value for this or not expose it at all. I can't think of a reason to have more than one polkit agent running at once, in which case we should just put it on a standard path for quickshell. If there is a reason to run more than one, we should still provide a default because most people wont want to or know how to configure this.
2. Formatting broke here

src/services/polkit/qml.hpp
```diff
@@ -0,0 +1,286 @@
...
+	/// The action ID that this authentication request is for.
+	Q_PROPERTY(QString actionId READ activeActionId NOTIFY authenticationRequestStarted);
+
+	/// The list of identities that may be used to authenticate.
+	///
+	/// Each identity may be a user or a group. You may select any of them to
+	/// authenticate by setting @@selectedIdentity. By default, the first identity
+	/// in the list is selected.
+	QSDOC_TYPE_OVERRIDE(ObjectModel<Identity>*)
+	Q_PROPERTY(UntypedObjectModel* identities READ activeIdentities NOTIFY authenticationRequestStarted);
```
outfoxxed: Models are best used on lists that are either very long or subject to change in a way that we'd like users to be able to handle individually (e.g. item removed from list animation). A simple QList<Identity*> should suffice here.

src/services/polkit/qml.hpp
```diff
@@ -0,0 +1,286 @@
...
+	/// Emitted after the identity that is authenticating has changed by user request.
+	void selectedIdentityChanged();
+
+	/// Emitted whenever the authentication conversation generates a new message to present.
+	///
+	/// This is always additional to the authentictation request's base message.
+	void subMessageChanged();
+
+	/// Emitted when the user needs to provide input to continue authentication.
+	void inputRequestChanged();
```
outfoxxed: Change signals don't need docstrings and are hidden by default

Cu3PO42:
Thanks for the review!

I don't like using a gobject library here either, but yes, there is a good reason for using it instead of interacting directly over DBus. PolKit requires the client that sends the confirmation message to be privileged. To this end, `libpolkit` has a setuid helper binary that it invokes, which handles the actual PAM conversation, and shuffles all of the data between the parent and the helper.

To avoid libpolkit, Quickshell would either need to run as root or we would need our own setuid helper. Besides the implementation overhead, this would break simply running it from your Flake via Nix or installing it via Home-Manager. I imagine (although I don't positively know) many distros might also look more sceptically at any setuid binaries.

Finally, the DBus interface isn't actually stable because it is currently considered an implementation detail. I don't think it has changed significantly in recent years, but I would still prefer the more stable solution.

outfoxxed:
That's horrible on so many levels.

Cu3PO42:
I... yeah. But that's Polkit for you. There's many things I wish I didn't know about it.

If, given these circumstances, you'd rather not have the code upstream, I'm just going to package it up as a separate library.

outfoxxed:
I'm fine with the gobject code since its an actual requirement, and this is a feature many people want. I'll finish the review tomorrow probably.

Cu3PO42:
Thank you! I'm not in any rush. Also, fair warning: this is the first 'not insignificant' thing I've done with Qt, so while I tried to do things right, some things might be the way they are because I didn't know any better. I'll happily take any pointers.

outfoxxed:
Thats fine. One further thing that would be helpful would be to add a manual test qml file similar to what exists in other parts of the source. Just cover a reasonable amount of the api for a simple manual regression test

src/services/polkit/qml.hpp
Cu3PO42: I simply wasn't aware of Q_GADGETs. I do currently use QObject's 'cleanup with parent' functionality, but I don't see why this couldn't easily be changed. Will do.

src/services/polkit/qml.hpp
Cu3PO42: 1. Fair enough. You actually can't have more than one PolKit agent running at the same time even if you choose different paths. I suppose I could just put `/org/quickshell/PolkitAgent` as the path?

Cu3PO42:
I got started on some of your comments. Here's an additional thing we might consider:

PolKit can send either 'show-info' or 'show-error'. I am currently storing both into the same property with an `isError` flag. I.e. an incoming error overwrites a previous info. PolKit itself does not specify the way the UI should look, but many agents handle it this way. However, for some scenarios showing both might make sense, e.g. fprintd will show 'Place your right index finger on the reader' as info and 'Fingerprint did not match' as info. I am considering having both individual properties and a combined one for the latest message so users can choose how to design their dialog.

One thing I'm a bit unsure on is object destruction/cleanup and its behavior in Qt. When an authentication request finishes, we need to clean up all objects, but does this cause the UI to flicker, if we animate a window out, for example?

src/services/polkit/qml.hpp
outfoxxed: whatever is the most standardish seeming path sounds good to me

src/services/polkit/qml.hpp
outfoxxed: FYI you can use Q_GADGETs as value types which don't need that at all

outfoxxed:
> PolKit can send either 'show-info' or 'show-error'. I am currently storing both into the same property with an isError flag. I.e. an incoming error overwrites a previous info. PolKit itself does not specify the way the UI should look, but many agents handle it this way. However, for some scenarios showing both might make sense, e.g. fprintd will show 'Place your right index finger on the reader' as info and 'Fingerprint did not match' as info. I am considering having both individual properties and a combined one for the latest message so users can choose how to design their dialog.

Does the api provide any sort of hint as to if both should be present or not? If it does something like sending a new version of the error message thats an empty string to reset thats a pretty clear signal we could have multiple, but if not I'm less sure. If the messages are coming from pam, our pam module currently replaces the last message with the new one so it probably makes sense to keep that design.

> One thing I'm a bit unsure on is object destruction/cleanup and its behavior in Qt. When an authentication request finishes, we need to clean up all objects, but does this cause the UI to flicker, if we animate a window out, for example?

This is a bit of a mess, Quickshell in particular has a hack for it called Retainable which you can see here https://quickshell.org/docs/v0.2.1/types/Quickshell/Retainable/. Retainable objects generally hold their last state and become immutable while retained to make it easier to write reactive UI against them.

Note: I tested the dialog and see you have a lot of `DEBUG: PolkitAgent: ` lines. Use logging categories (qCDebug etc) for this. We hide them by default but they can be as detailed as you want (though you should make an effort to not put passwords in them)

Cu3PO42:
> Does the api provide any sort of hint as to if both should be present or not? If it does something like sending a new version of the error message thats an empty string to reset thats a pretty clear signal we could have multiple, but if not I'm less sure. If the messages are coming from pam, our pam module currently replaces the last message with the new one so it probably makes sense to keep that design.

Unfortunately not. The messages are indeed forwarded from PAM. I am therefore inclined to leave it as is.

> This is a bit of a mess, Quickshell in particular has a hack for it called Retainable which you can see here https://quickshell.org/docs/v0.2.1/types/Quickshell/Retainable/. Retainable objects generally hold their last state and become immutable while retained to make it easier to write reactive UI against them.

Thanks! I'll look into that to figure out how best to use it in this scenario.

> Note: I tested the dialog and see you have a lot of DEBUG: PolkitAgent: lines. Use logging categories (qCDebug etc) for this. We hide them by default but they can be as detailed as you want (though you should make an effort to not put passwords in them)

That's exactly what I was looking for. I'll migrate to that.

src/services/polkit/listener.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+bool qs_polkit_agent_register(QsPolkitAgent* agent) {
+	if (agent->agent->path().isEmpty()) {
+		qCWarning(logPolkitListener) << "cannot register listener without a path set.";
+		return false;
+	}
+
+	GError* error = nullptr;
+	auto subject = polkit_unix_session_new_for_process_sync(getpid(), nullptr, &error);
```
outfoxxed: If this involves remote ipc we probably shouldn't be using the sync variant. The glib event loop should be backing the qt one on linux, so callbacks in theory would execute here.

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,428 @@
...
+bool InputRequest::echo() const { return mEcho; }
+
+static std::unordered_map<QString, PolkitAgent*> registeredAgentsByPath {};
+static std::unordered_map<QString, PolkitAgent*> waitingAgentsByPath {};
+
+PolkitAgent::PolkitAgent(QObject* parent): QObject(parent), listener(qs_polkit_agent_new(this)) {}
+
+PolkitAgent::~PolkitAgent() {
+	// First, destroy all but the first request in the queue, so that the cancel
+	// method doesn't start a new one.
```
outfoxxed: I would think its not unreasonable to handle a request right after cancelling a different one, especially if they're for different programs.

Also, when reloading the graph, preserving the old polkit session may be the most reasonable case, especially if the user wants to do something like edit their auth dialog with a real app running to test with.

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,428 @@
...
+	g_object_unref(listener);
+
+	if (auto it = registeredAgentsByPath.find(mPath);
+	    it != registeredAgentsByPath.end() && it->second == this)
+	{
+		registeredAgentsByPath.erase(mPath);
+
+		// In case of config reloads new objects are constructed before the old ones
+		// are destroyed. Therefore the new agent cannot take over the path still in
+		// use by the old agent.
```
outfoxxed: It can as long as its negotiated. I think we established it doesn't make sense to have more than one agent, so a single agent per process solves this nicely.

More than one polkit agent object at a time in a single generation however does not make sense and should print an error message and render any agent objects >1 inoperable.

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,428 @@
...
+	}
+	return queuedRequests.front()->iconName;
+}
+
+const QString& PolkitAgent::activeActionId() const {
+	if (queuedRequests.empty()) {
+		return emptyString;
+	}
+	return queuedRequests.front()->actionId;
+}
```
outfoxxed: single line if when its short

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,428 @@
...
+		Identity* obj;
+
+		if (POLKIT_IS_UNIX_USER(identity)) {
+			auto uid = polkit_unix_user_get_uid(POLKIT_UNIX_USER(identity));
+			auto pw = getpwuid(uid);
+			auto name = (pw && pw->pw_name && *pw->pw_name) ? QString::fromUtf8(pw->pw_name)
+			                                                : QString::number(uid);
+			QString icon;
+			if (pw && pw->pw_dir && *pw->pw_dir) {
+				icon = QString::fromUtf8(pw->pw_dir) + QDir::separator() + ".face.icon";
```
outfoxxed: Is this a standard? I've heard of multiple paths to a user picture

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+		// For now, other identity types (currently only PolkitUnixNetgroup)
+		// are not supported.
+
+		if (obj) {
+			mIdentities.append(obj);
+		}
+	}
+
+	mSelectedIdentity = mIdentities.isEmpty() ? nullptr : mIdentities.first();
```
outfoxxed: I'm not sure how much significance it has with polkit, but would it make sense for a user to pick any other identity if multiple are available?

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+	qCDebug(logPolkit) << "setting up authentication session for identity"
+	         << (mSelectedIdentity ? mSelectedIdentity->name() : "<null>");
+
+	currentSession = new Session(mSelectedIdentity->polkitIdentity, req.cookie);
+
+	connect(currentSession, SIGNAL(request(QString, bool)), this, SLOT(request(QString, bool)));
+	connect(currentSession, SIGNAL(completed(bool)), this, SLOT(completed(bool)));
+	connect(currentSession, SIGNAL(showError(QString)), this, SLOT(showError(QString)));
+	connect(currentSession, SIGNAL(showInfo(QString)), this, SLOT(showInfo(QString)));
```
outfoxxed: Don't use SIGNAL and SLOT macros they're a legacy api from Qt 4. Use the format that takes member pointers. Also prefix static functions with their class (QObject::connect)

src/services/polkit/qml.cpp
Cu3PO42: Unfortunately there is no universally agreed upon standard. This is one of the conventions. The overall most robust way is probably to ask AccountsService over DBus for the path, which is where Gnome and modern versions of KDE store it. (* SDDM actually doesn't query AccountsService, it simply looks in the default `/var/lib/AccountsService/` path for the user icon regardless of what you actually set. This breaks on NixOS, for example :upside_down_face: )

The reason why I went with this very simple approach is that it's very easy for users to supply their own picture, if they want to. I would expect users of Quickshell on WM-like-compositors to not necessarily have any tool to set the icon in AccountsService, whereas creating a file is trivial. Noctalia actually uses `~/.face`.

That said, I have no strong opinion on this and we could simply leave it out due to lack of a standard.

src/services/polkit/qml.cpp
Cu3PO42: It might in some circumstances. For example, the user may be logged in as a user that is not in `wheel`/whatever the specific policy is configured to require, but have the credentials to a different account that has the necessary permissions.

In this case, they could authenticate as a different user.

src/services/polkit/qml.cpp
Cu3PO42: Handling one request after another is the default behavior in `finishAuthenticationRequest`. The `cancelAuthenticationRequest` method calls `finishAuthenticationRequest` and therefore would start another request immediately.

However, we are in the destructor of the agent here, so there will be no-one to handle the queued requests after the destructor has run. Therefore I want to reject all currently queued requests and not start processing them. To this end, I start by rejecting all requests after the current one, so that when the cancel method runs, it won't have any more to start.

In the case of reloading, I agree, preservation would be ideal. I'll look into how I can solve that cleanly only for the case of the reload, but still dismiss everything when actually destroying the agent.

src/services/polkit/qml.cpp
Cu3PO42: > It can as long as its negotiated.

Are you referring to some particular Quickshell interface or is this a general remark? If the former, do you have a pointer? If the latter, this code is the early version of this negotiation. I meant to express 'it cannot take over the path without doing something to support it explicitly'. In the old version using polkit-qt, it wasn't possible to transfer the listener over, so closing the path and then opening it again was necessary. Now it might be possible and cleaner to transfer the listener. I'll look into it.

src/services/polkit/listener.cpp
Cu3PO42: After looking into the implementation and confirming that there are DBus calls involved, I switched to the async API. I wondered before why callbacks from glib would 'just working', I did not realize Qt based their event loop on top of glib on Linux. Thanks for sharing that!

src/services/polkit/qml.cpp
outfoxxed: Then we should probably expose that as an option

src/services/polkit/qml.cpp
outfoxxed: I would think to leave it out, at least for now, since it isn't really following a standard then.

src/services/polkit/qml.cpp
outfoxxed: I'm not sure I fully understand the particular problem here, but singleton resources that live longer than a single engine generation are used in a few places in Quickshell.

Since the qml engine controls the lifecycle of all objects created in it, Quickshell often has an internal singleton object with a thin wrapper object owned by the qml engine instead, so it doesn't neeto die with the generation.

With the restriction that we can only reasonably have one polkit agent at once, it seems logical to me to represent it this way.

What I meant by negotiation was only allowing one qml owner of the polkit agent per generation, but keeping tha actual agent object the same.

src/services/polkit/qml.cpp
outfoxxed: As for reloading see my comment below about using an internal singleton. you can wait for the destruction of the last engine generation with a ref to the backing agent to cancel all requests and deregister the agent, or some similar pattern.

src/services/polkit/qml.cpp
Cu3PO42: The user can set `selectedIdentity` at any time, even after authentication is already started. This is just a (in my opinion) reasonable default. The alternative would be to not set this and always make the user choose.

This behavior also mirrors what all existing PolKit agents that I'm aware of do.

src/services/polkit/qml.cpp
Cu3PO42: I think we agree conceptually, I'm just not sure of the exact API surface that permits me to do this cleanly. I can obviously put the listener itself into a singleton and introduce a thin wrapper. Since we want to permit only one such wrapper per generation, I need to add a check to this effect: If a new `PolkitAgent` is constructed, I can check this singleton and see one already exists and set an error state.

In the case of a config reload, I will also see the listener is already in use, but in this case I want to take ownership of it. I could attempt to use `currentGeneration` to detect this, but this seems mildly heuristic. Alternatively, I could extend `Reloadable` and do it in `Reloadable#onReload`. Do you have a preference for any of these approaches or is there some other idiomatic way I'm not seeing currently?

Thanks!

outfoxxed: Theres a couple differences in the approaches you could take:
- Reloadable uses heuristics to try to match an object in the old generation to one in the new generation. This is the system that allows allocated windows to be reused. You could use this if you wanted to only persist the polkit agent if the user placed it in the same spot in the same file across generations.
- PostReloadHook fires unconditionally after Reloadable and has no concept of the scene graph. Combining this with `findObjectGeneration` on the user facing wrapper objects from the current and previous generations will tell you if we're switching generations or the user created a second invalid agent. Avoid `currentGeneration`.

Either of these interfaces you would implement on the wrapper object and not the internal singleton.

I recommend the PostReloadHook approach but would accept either.

src/services/polkit/qml.cpp
outfoxxed: I missed the interface for changing it. That behavior is what I'd expect.

Cu3PO42:
I performed a fairly significant refactoring to accommodate the changes you recommended.

* `PolkitAgent` is now a very thin wrapper with the business logic living in `PolkitAgentImpl`.
* `PostReloadHook` is used to handle transitions between generations.
* State of the currently active request as well as any queued requests is preserved on reload.
* `AuthFlow` is a new class that houses the actual state of an in-progress authentication. It is `Retainable` to accommodate animating out dialogs.
* The code is just better organized overall.

I'll look at the lint errors soon.

src/services/polkit/flow.hpp
```diff
@@ -0,0 +1,162 @@
...
+	QML_UNCREATABLE("AuthFlow can only be obtained from PolkitAgent.");
+
+	// clang-format off
+	/// The main message to present to the user.
+	Q_PROPERTY(QString message READ message CONSTANT);
+
+	/// The icon to present to the user in association with the message.
+	///
+	/// The icon name follows the [FreeDesktop icon naming specification](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html).
+	Q_PROPERTY(QString iconName READ iconName CONSTANT);
```
outfoxxed, line 31: Best to link to @@Quickshell.Quickshell.iconPath() to display it in an image

src/services/polkit/flow.hpp
```diff
@@ -0,0 +1,162 @@
...
+	///
+	/// Each identity may be a user or a group. You may select any of them to
+	/// authenticate by setting @@selectedIdentity. By default, the first identity
+	/// in the list is selected.
+	Q_PROPERTY(QList<Identity*> identities READ identities CONSTANT);
+
+	/// The identity that will be used to authenticate.
+	/// 
+	/// Setting this will abort any ongoing authentication conversations and start a new one.
+	Q_PROPERTY(Identity* selectedIdentity READ selectedIdentity WRITE setSelectedIdentity NOTIFY selectedIdentityChanged);
```
outfoxxed: "Setting" -> "Changing"

Reactive changes can cause a lot of "Setting" so this wording could be confusing

src/services/polkit/flow.hpp
```diff
@@ -0,0 +1,162 @@
...
+	/// This message is used to prompt the user for required input.
+	Q_PROPERTY(QString inputPrompt READ inputPrompt NOTIFY responseRequestChanged);
+
+	/// Indicates whether the user's response should be visible. (e.g. for passwords this should be false)
+	Q_PROPERTY(bool responseVisible READ responseVisible NOTIFY responseRequestChanged);
+
+	/// An additional message to present to the user.
+	///
+	/// This may be used to show errors or supplementary information.
+	Q_PROPERTY(QString supplementaryMessage READ supplementaryMessage NOTIFY supplementaryChanged);
```
outfoxxed: Add something along the lines of "See @@supplementaryIsError to determine if this should be displayed as an error message"

src/services/polkit/flow.cpp
```diff
@@ -0,0 +1,159 @@
...
+	qCDebug(logPolkitState) << "authentication session completed, gainedAuthorization ="
+	                        << gainedAuthorization << ", isCancelled =" << mIsCancelled;
+
+	if (gainedAuthorization) {
+		mIsCompleted = true;
+		mIsSuccessful = true;
+		mRequest->complete();
+
+		emit completedChanged();
+		emit authenticationSucceeded();
```
outfoxxed: this-> on method calls including emit, as well as members. ci should be catching this one if its not

src/services/polkit/flow.hpp
```diff
@@ -0,0 +1,162 @@
...
+	QList<Identity*> mIdentities;
+	Identity* mSelectedIdentity = nullptr;
+	bool mIsResponseRequired = false;
+	QString mResponseMessage;
+	bool mResponseVisible = false;
+	QString mExtraMessage;
+	bool mExtraIsError = false;
+	bool mIsCompleted = false;
+	bool mIsSuccessful = false;
+	bool mIsCancelled = false;
```
outfoxxed: While acceptable as-is, properties like this are better handled with QObject bindable properties, which are basically C++-side reactive properties.

src/services/polkit/flow.cpp
```diff
@@ -0,0 +1,159 @@
...
+const QString& AuthFlow::iconName() const { return mRequest->iconName; }
+const QString& AuthFlow::actionId() const { return mRequest->actionId; }
+const QString& AuthFlow::cookie() const { return mRequest->cookie; }
+const QList<Identity*>& AuthFlow::identities() const { return mIdentities; }
+
+Identity* AuthFlow::selectedIdentity() const { return mSelectedIdentity; }
+
+void AuthFlow::setSelectedIdentity(Identity* identity) {
+	if (mSelectedIdentity == identity) return;
+	if (!identity) return; // ignore null changes
```
outfoxxed: I would print an error here with qmlWarning(this)

src/services/polkit/flow.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+namespace qs::service::polkit {
+AuthFlow::AuthFlow(AuthRequest* request, QList<Identity*>&& identities, QObject* parent)
+    : QObject(parent)
+    , mRequest(request)
+    , mIdentities(std::move(identities)) {
+	mSelectedIdentity = mIdentities.isEmpty() ? nullptr : mIdentities.first();
+	if (!mSelectedIdentity) {
+		qCCritical(logPolkitState) << "AuthFlow created with no valid identities!";
```
outfoxxed: Since this is internal and should never happen, i would make this a qCFatal, and reject this case if polkit ever tries to send it, before reaching the user.

src/services/polkit/flow.cpp
```diff
@@ -0,0 +1,159 @@
...
+	} else if (mIsCancelled) {
+		mIsCompleted = true;
+		mIsSuccessful = false;
+
+		emit completedChanged();
+	} else {
+		emit authenticationFailed();
+
+		clearState();
+		setupSession();
```
outfoxxed: Creating a new session to try again feels very weird but I'm going to assume this is just polkit insanity

src/services/polkit/identity.hpp
```diff
@@ -0,0 +1,59 @@
...
+typedef struct _PolkitIdentity PolkitIdentity;
+
+namespace qs::service::polkit {
+//! Represents a user or group that can be used to authenticate.
+class Identity: public QObject {
+	Q_OBJECT;
+
+	// clang-format off
+	/// The Id of the identity. If the identity is a user, this is the user's uid. See @@isGroup.
+	Q_PROPERTY(id_t id READ id CONSTANT);
```
outfoxxed: Please cast/rename to a qt primitive type in the interface. This will show up as `unknown` in the doc generator.

src/services/polkit/identity.cpp
```diff
@@ -0,0 +1,62 @@
...
+    : QObject(parent)
+    , polkitIdentity(polkitIdentity)
+    , mId(id)
+    , mName(std::move(name))
+    , mDisplayName(std::move(displayName))
+    , mIsGroup(isGroup) {}
+
+Identity::~Identity() = default;
+
+Identity* Identity::fromPolkitIdentity(PolkitIdentity* identity) {
```
outfoxxed: I don't see identities ever being deleted after creation, in agentimpl, flow, or otherwise. I suggest using the qobject parent/child hierarchy or making this a Q_GADGET.

src/services/polkit/listener.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+static void qs_polkit_agent_register_cb(GObject*, GAsyncResult* res, gpointer userData);
+void qs_polkit_agent_register(QsPolkitAgent* agent, const char* path) {
+	if (path == nullptr || path[0] == '\0') {
+		qCWarning(logPolkitListener) << "cannot register listener without a path set.";
+		agent->cb->registerComplete(false);
+		return;
+	}
+
+	auto data = new register_cb_data {.agent = agent, .path = strdup(path)};
```
outfoxxed: 1. Do we need gobject ref/unref on the polkit agent?
2. Can we use std::string instead of passing more raw cstrings around?

src/services/polkit/listener.cpp
```diff
@@ -0,0 +1,202 @@
...
+	// Identities may be duplicated, so we use the hash to filter them out.
+	std::unordered_set<guint> identitySet;
+	std::vector<PolkitIdentity*> identityVector;
+	for (auto item = g_list_first(identities); item != nullptr; item = g_list_next(item)) {
+		auto identity = static_cast<PolkitIdentity*>(item->data);
+		if (identitySet.contains(polkit_identity_hash(identity))) continue;
+
+		identitySet.insert(polkit_identity_hash(identity));
+		identityVector.push_back(identity);
+		g_object_ref(identity);
```
outfoxxed: I see we add a ref to the identities we're using. Add a comment if polkit takes care of derefing all the ones we dont' or do it if we need to

src/services/polkit/listener.cpp
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	auto request = new qs::service::polkit::AuthRequest {
+	    .actionId = QString::fromUtf8(actionId),
+	    .message = QString::fromUtf8(message),
+	    .iconName = QString::fromUtf8(iconName),
+	    .cookie = QString::fromUtf8(cookie),
```
outfoxxed: Comment on these string lifetimes as well, should we be freeing them?

src/services/polkit/listener.cpp
```diff
@@ -0,0 +1,202 @@
...
+) {
+	auto self = QS_POLKIT_AGENT(listener);
+
+	auto asyncResult = g_task_new(reinterpret_cast<GObject*>(self), nullptr, callback, userData);
+
+	// Identities may be duplicated, so we use the hash to filter them out.
+	std::unordered_set<guint> identitySet;
+	std::vector<PolkitIdentity*> identityVector;
+	for (auto item = g_list_first(identities); item != nullptr; item = g_list_next(item)) {
+		auto identity = static_cast<PolkitIdentity*>(item->data);
```
outfoxxed: It might be a good idea to have an RAII wrapper around these so you don't have to reason about when to unref them. Also if you use QList its internally shared which looks like something we want here.

src/services/polkit/qml.hpp
```diff
@@ -0,0 +1,82 @@
...
+    /// The D-Bus path that this agent listener will use.
+	///
+	/// If not set, a default of /org/quickshell/Polkit will be used.
+    Q_PROPERTY(QString path READ path WRITE setPath);
+
+	/// Indicates whether the agent registered successfully and is in use.
+	Q_PROPERTY(bool isRegistered READ isRegistered NOTIFY isRegisteredChanged);
+
+	/// Indicates an ongoing authentication request.
+	///
```
outfoxxed, line 42: formatting broken. You should leave clang-format on unless its breaking line wrapping

src/services/polkit/qml.hpp
```diff
@@ -0,0 +1,82 @@
...
+	/// If this is true, other properties such as @@message and @@iconName will
+	/// also be populated with relevant information.
+	Q_PROPERTY(bool isActive READ isActive NOTIFY isActiveChanged);
+
+	/// The current authentication state if an authentication request is active.
+	/// Null when no authentication request is active.
+	Q_PROPERTY(AuthFlow* flow READ flow NOTIFY flowChanged);
```
outfoxxed: The idea behind authflow being a separate object is that you can retain it right? Not a criticism just making sure

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+namespace qs::service::polkit {
+PolkitAgent::PolkitAgent(QObject* parent): PostReloadHook(parent) {}
+
+PolkitAgent::~PolkitAgent() = default;
+
+void PolkitAgent::classBegin() {
+	// Nothing to do here.
+}
```
outfoxxed: just leave classbegin () {} in the header

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "../../core/logcat.hpp"
+
+namespace {
+QS_LOGGING_CATEGORY(logPolkit, "quickshell.service.polkit");
+}
+
+namespace qs::service::polkit {
+PolkitAgent::PolkitAgent(QObject* parent): PostReloadHook(parent) {}
+
+PolkitAgent::~PolkitAgent() = default;
```
outfoxxed: this can also go in the header

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,69 @@
...
+void PolkitAgent::classBegin() {
+	// Nothing to do here.
+}
+
+void PolkitAgent::componentComplete() {
+	PostReloadHook::componentComplete();
+
+	if (mPath.isEmpty()) mPath = "/org/quickshell/Polkit";
+
+	PolkitAgentImpl::tryGetOrCreate(this);
```
outfoxxed: It seems like a better answer here is to make this tryTakeover and drop postreloadhook, otherwise the old generation's agent will be destroyed before this one takes over, which would lead to us dropping sessions

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return nullptr;
+}
+
+void PolkitAgent::onPostReload() {
+	if (!PolkitAgentImpl::tryTakeover(this)) return;
+
+	emit isRegisteredChanged();
+	emit isActiveChanged();
+	emit flowChanged();
```
outfoxxed: Another good site for bindable properties

src/services/polkit/agentimpl.hpp
```diff
@@ -0,0 +1,49 @@
...
+	/// Finalize and remove the current authentication request.
+	void finishAuthenticationRequest();
+
+	QsPolkitAgent* listener = nullptr;
+	bool isRegistered = false;
+
+	PolkitAgent* qmlAgent = nullptr;
+
+	AuthFlow* activeFlow = nullptr;
+	std::deque<AuthRequest*> queuedRequests;
```
outfoxxed, line 59: It seems like we're accepting every request preemptively and then keeping them waiting. If we don't do that does polkit queue them or cancel them? Should we be leaving that to polkit?

src/services/polkit/flow.cpp
Cu3PO42: I actually do reject any such requests before creating an `AuthFlow` in `PolkitAgentImpl`. This is just a sanity check that the flow object is being constructed correctly. Nevertheless I changed it to fatal.

src/services/polkit/flow.cpp
Cu3PO42: Indeed. This is just the way the library works.

src/services/polkit/identity.cpp
Cu3PO42: I thought I set a parent for them, but I didn't previously. I made them a child of the AuthFlow to take care of this.

src/services/polkit/listener.cpp
Cu3PO42: 1. I don't think we strictly do, since the caller mustn't unref while an async operation is running. However, I did it just in case since it can't hurt.
2. We can and I did.

src/services/polkit/qml.hpp
Cu3PO42: Yup! The way I understand it, this should allow to animate one request out while the next is animating in, for example.

src/services/polkit/qml.cpp
Cu3PO42: This must be a case of not seeing the forest for the trees. This is indeed much simpler and I changed it. That said, the post reload hook always run before the destructor of the previous generation in my testing.

src/services/polkit/agentimpl.hpp
Cu3PO42: Polkit pushes requests to us as soon as they are created by some client. We need to handle them in some form:

- reject them immediately.
- queue them as I do here.
- drop them and let them timeout.

Of these options I like the queue version best.

Cu3PO42:
Thank you for the extensive review! I addressed most of your comments. I also introduced a RAII helper `GObjectRef` to make handling of GObject objects less error prone. I still need to look into bindable properties.

I selectively turned off some lints in specific instances where I felt they were not helpful or actively wrong (e.g. it won't let me forward declare any structs from the Polkit libraries). There should always be an accompanying comment explaining why the specific rule doesn't make sense in that case. Let me know if you disagree with any of my reasons.

src/services/polkit/qml.cpp
Cu3PO42: ~~Based on my current understanding of bindable properties (at least as far as the `Q_OBJECT_BINDABLE_PROPERTY`) is concerned, are backed by a field in the object. This is not the case for the properties that are associated with these signals. I am sure they can be hooked up to work anyway, but I'm not sure if that would still be simpler?~~

I think I've figured it out. Will hopefully push something later.

outfoxxed: They are field backed but when used with a binding callback the field acts as more of a cache than anything else. It also compares old/new values to see if anything changed and emit signals or not.

There are cases where it improves readability, generally the more values that depend on eachother the greater the improvement.

Here's an example of it being used with a binding: https://github.com/quickshell-mirror/quickshell/blob/3e2ce40b18af943f9ba370ed73565e9f487663ef/src/wayland/wlr_layershell/wlr_layershell.cpp#L18

And heres an example of it being used as a field that just so happens to emit signals automatically on prop change: https://github.com/quickshell-mirror/quickshell/blob/3e2ce40b18af943f9ba370ed73565e9f487663ef/src/services/notifications/notification.cpp#L113

Up to you where you use it but it does generally make things clearer.

Cu3PO42:
I do love a good reactivity system. I switched over to bindable properties where I figured it makes sense. With that I believe I have addressed everything you pointed out. I appreciate your patience, this has been quite the learning experience.

Cu3PO42:
I rebased onto current master, squashed and fixed a small bug in the reloading logic.

I have also integrated this into Noctalia [in a branch](https://github.com/Cu3PO42/noctalia-shell/tree/polkit-agent) to validate in a more complex config.

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace {
+QS_LOGGING_CATEGORY(logPolkit, "quickshell.service.polkit");
+}
+
+namespace qs::service::polkit {
+PolkitAgent::PolkitAgent(QObject* parent): QObject(parent) {
+	// Try to takeover an existing PolkitAgentImpl from the previous QML engine generation.
+	// If none exists, we wait until componentComplete to create one so we have the correct path.
+	PolkitAgentImpl::tryTakeover(this);
+}
```
outfoxxed: Why not just wait until component complete and takeoverOrCreate then?

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,49 @@
...
+	if (this->mPath.isEmpty()) this->mPath = "/org/quickshell/Polkit";
+
+	auto* impl = PolkitAgentImpl::tryGetOrCreate(this);
+	if (impl == nullptr) return;
+
+	impl->setPath(this->mPath);
+
+	this->bFlow.setBinding([impl]() -> AuthFlow* { return impl->activeFlow().value(); });
+	this->bIsActive.setBinding([impl]() -> bool { return impl->activeFlow().value() != nullptr; });
+	this->bIsRegistered.setBinding([impl]() -> bool { return impl->isRegistered().value(); });
```
outfoxxed: Do you need the return types here or can it just be `[impl] {}`?

src/services/polkit/qml.cpp
```diff
@@ -0,0 +1,49 @@
...
+	if (this->mPath.isEmpty()) {
+		this->mPath = path;
+	} else if (this->mPath != path) {
+		qCWarning(logPolkit) << "cannot change path after it has been set.";
+	}
+}
+
+QBindable<AuthFlow*> PolkitAgent::flow() { return &this->bFlow; }
+QBindable<bool> PolkitAgent::isActive() { return &this->bIsActive; }
+QBindable<bool> PolkitAgent::isRegistered() { return &this->bIsRegistered; }
```
outfoxxed: These can go in the header

src/services/polkit/qml.hpp
```diff
@@ -0,0 +1,89 @@
...
+	/// Indicates an ongoing authentication request.
+	///
+	/// If this is true, other properties such as @@message and @@iconName will
+	/// also be populated with relevant information.
+	Q_PROPERTY(bool isActive READ default NOTIFY isActiveChanged BINDABLE isActive);
+
+	/// The current authentication state if an authentication request is active.
+	/// Null when no authentication request is active.
+	Q_PROPERTY(AuthFlow* flow READ default NOTIFY flowChanged BINDABLE flow);
+	// clang-format on
```
outfoxxed: Comment formatting broken, and it looks like you dont need clang-format off for this class.

src/services/polkit/flow.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+QBindable<bool> AuthFlow::isResponseRequired() { return &this->bIsResponseRequired; }
+QBindable<QString> AuthFlow::inputPrompt() { return &this->bInputPrompt; }
+QBindable<bool> AuthFlow::responseVisible() { return &this->bResponseVisible; }
+QBindable<QString> AuthFlow::supplementaryMessage() { return &this->bSupplementaryMessage; }
+QBindable<bool> AuthFlow::supplementaryIsError() { return &this->bSupplementaryIsError; }
+QBindable<bool> AuthFlow::isCompleted() { return &this->bIsCompleted; }
+QBindable<bool> AuthFlow::isSuccessful() { return &this->bIsSuccessful; }
+QBindable<bool> AuthFlow::isCancelled() { return &this->bIsCancelled; }
+AuthRequest* AuthFlow::authRequest() const { return this->mRequest; }
```
outfoxxed: Bindables inline in header

src/services/polkit/flow.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+void AuthFlow::submit(const QString& value) {
+	if (!this->currentSession) return;
+
+	qCDebug(logPolkitState) << "submitting response to authentication request";
+
+	this->currentSession->respond(value);
+
+	this->bIsResponseRequired = false;
+	this->bInputPrompt = QString {};
```
outfoxxed: QString() constructor

src/services/polkit/flow.hpp
```diff
@@ -0,0 +1,193 @@
...
+	);
+	Q_OBJECT_BINDABLE_PROPERTY(
+	    AuthFlow,
+	    bool,
+	    bSupplementaryIsError,
+	    &AuthFlow::supplementaryIsErrorChanged
+	);
+	Q_OBJECT_BINDABLE_PROPERTY(AuthFlow, bool, bIsCompleted, &AuthFlow::isCompletedChanged);
+	Q_OBJECT_BINDABLE_PROPERTY(AuthFlow, bool, bIsSuccessful, &AuthFlow::isSuccessfulChanged);
+	Q_OBJECT_BINDABLE_PROPERTY(AuthFlow, bool, bIsCancelled, &AuthFlow::isCancelledChanged);
```
outfoxxed: clang-format off here please to keep these inline

src/services/polkit/gobjectref.hpp
```diff
@@ -0,0 +1,65 @@
...
+	explicit GObjectRef(T* ptr, GObjectNoRefTag /*tag*/): ptr(ptr) {}
+
+	~GObjectRef() {
+		if (this->ptr) {
+			g_object_unref(this->ptr);
+		}
+	}
+
+	// We do handle self-assignment in a more general case by checking the
+	// included pointers rather than the wrapper objects themselves.
```
outfoxxed, line 28: The comment says this which makes sense but it looks like we dont ever compare the gobjectref content pointers before unreffing?

src/services/polkit/qml.cpp
Cu3PO42: While integrating into Noctalia, I ran into a situation where the destructor of the old generation was called before componentComplete of the next one was, leading to resource destruction. It may very well be that this was due to the way Noctalia handle reloads (where I e.g. also observed the bar disappearing and reappearing), but in any case I moved this as early as possible to avoid rejected requests wherever possible.

Is it true that (if nothing else is going on) `componentComplete` of the next generation is always called before the destructor of the previous is called? If yes, I can indeed simplify this.

src/services/polkit/gobjectref.hpp
Cu3PO42: I overloaded `operator==` to compare the content pointers. Therefore, this check is in the first line of the both assignment operators. The lint rule would like me to check `this == &other` instead of `*this == other`, which would be less general. I could expand the comment to make this more clear.

outfoxxed: Whoops yeah. My misread.

src/services/polkit/qml.cpp
outfoxxed: componentComplete absolutely *should* be trigged before destruction of the next generation. We can't rely on property values until then so onReload needs to wait for it before destroying the old generation. If that is not the case, Noctilla probably delays things with loaders in a way that breaks the reload system. @ly-sec?

ref
https://github.com/quickshell-mirror/quickshell/blob/1b147a2c78983877909f9e531fc8ce17c35a297a/src/core/generation.cpp#L144
https://github.com/quickshell-mirror/quickshell/blob/1b147a2c78983877909f9e531fc8ce17c35a297a/src/core/rootwrapper.cpp#L136-L143

Cu3PO42:
Minor style question: I have moved the bindable getters to the header, should I do the same for other property getters as well? Other than that I believe I got to everything. I've left it as separate commits for now to make it easier to review, but I can squash later.

outfoxxed:
Generally my policy is one-line getter functions can go in the header. Takeover change looks fine here.

Cu3PO42:
Alright, thanks! I went through and moved all one-line getters to the headers.

outfoxxed (changes requested):
Test issues discovered:
1. Polkit agent does not attempt to register itself if it failed and then the currently active agent leaves the bus. See notifications.
2. Debug log levels are still enabled
3. Supplementary error message is never populated for bad passwords

Also noticed: for bindables use Qt::beginPropertyUpdateGroup / end update group to not send useless updates.

Cu3PO42:
> Test issues discovered:
> 
>     1. Polkit agent does not attempt to register itself if it failed and then the currently active agent leaves the bus. See notifications.

This is true. I didn't tackle that before since it seemed to add significant complexity for relatively little advantage. However, I see the `QDbusServiceWatcher` actually makes this quite simple. Will do.

>     2. Debug log levels are still enabled

My bad. I misunderstood how the logging categories worked. I thought the default log level was a runtime switch, not in the code.

>     3. Supplementary error message is never populated for bad passwords

True. That was an oversight when designing `clearState`. It never came up for me in testing since my PAM config starts with fingerprints and the `show-info` from that overwrites any error message on restart anyway. I'll keep the supplementary message in that case. (Side note: I really wish PAM had better structuring to conversations.)


> Also noticed: for bindables use Qt::beginPropertyUpdateGroup / end update group to not send useless updates.

Aha! I was looking for something like that, thanks!

outfoxxed:
For the log levels just add QtWarningMsg to the logging category declaration to set the default. QT_LOGGING_RULES is then usable as a runtime switch.

Cu3PO42:
I have looked into registering once the active agent unregisters. Unfortunately, there is no specified bus names that agents need to register. They simply need to have an object on the system bus implementing the agent interface. Since agents generally deny any introspection request on the system bus, we cannot reliably determine which object or service we need to wait for.

Also, the Polkit authority does not provide any methods for obtaining the active agent, neither over D-Bus nor via other APIs. At best, we could come up with some heuristics based on running processes or simple timeouts. At that point, I would seriously question whether that is reasonable and worth it.

outfoxxed:
At that point it just isn't worth it for an edga case. No problem leaving it off.

Cu3PO42:
That should be everything. I have removed the entire `clearState` method in `Flow` intentionally. It previously reset the supplementary message and input request. We want to explicitly keep the supplementary message; and input request is already reset in `submit`, which always comes before `clearState` was called in `onComplete`.

outfoxxed:
LGTM overall. I'm assuming prompts like this in the qt polkit agent are created just from determining auth failed and not from a pam message.
image

I was able to recreate this with the authenticationFailed signal, but I think it would be helpful to add a property signifying if the flow has had a failed attempt instead of forcing users to connect to the signal. I've taken the liberty of adding this under the assumption that I'm correct, feel free to revert that if its not.

Ready to merge this. Thanks for dealing with all the reviews!

outfoxxed (approved):

Cu3PO42:
I do believe you are correct, this message is just derived from a failed session and the property seems a perfectly reasonable way to expose that. Thanks for adding that!

I've learned quite a lot about Qt doing this, so thank you for your tips and directions as well! Getting this merged is a nice birthday present :)

PR #356, opened 2025-11-11
i3/ipc: implement reception of arbitrary IPC events in QML

3akev:
This PR exposes the [subscribe](https://i3wm.org/docs/ipc.html#_subscribing_to_events) call to the QML API, allowing clients to receive subscribed events directly through rawEvent.

My use case is to receive input events in order to detect keyboard layout changes. See [here](https://github.com/3akev/noctalia-shell/blob/b2726a57abd792f04d3219cbde69acb97b8c3ab4/Services/Compositor/SwayService.qml#L41) and [here](https://github.com/3akev/noctalia-shell/blob/b2726a57abd792f04d3219cbde69acb97b8c3ab4/Services/Compositor/SwayService.qml#L321-L332) for an example doing this.

### DIFF
--- a/src/x11/i3/ipc/connection.cpp
+++ b/src/x11/i3/ipc/connection.cpp
@@ -1,4 +1,4 @@
-#include <algorithm>
+#include "connection.hpp"
 #include <array>
 #include <cstring>
 #include <tuple>
@@ -23,11 +23,6 @@
 #include <qtypes.h>
 
 #include "../../../core/logcat.hpp"
-#include "../../../core/model.hpp"
-#include "../../../core/qmlscreen.hpp"
-#include "connection.hpp"
-#include "monitor.hpp"
-#include "workspace.hpp"
 
 namespace qs::i3::ipc {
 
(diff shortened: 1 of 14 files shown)


outfoxxed:
The intention here is fine but it probably shouldn't be implemented like this, because subscriptions are not dropped between reloads. I would prefer to see something more declarative, e.g. `I3IpcListener { subscriptions: []; onMessage: ... }` unless you have a particular reason to do it this way.

3akev:
Implemented a new QML type, as suggested. [Here's a usage example.](https://github.com/3akev/noctalia-shell/blob/04939965f6c51e15edcc94bb33dc24fefb846869/Services/Compositor/SwayService.qml#L331-L336)  Let me know if further changes are needed.

src/x11/i3/ipc/connection.cpp
```diff
@@ -74,52 +81,67 @@ I3Ipc::I3Ipc() {
...
 	// clang-format off
-	QObject::connect(&this->liveEventSocket, &QLocalSocket::errorOccurred, this, &I3Ipc::eventSocketError);
-	QObject::connect(&this->liveEventSocket, &QLocalSocket::stateChanged, this, &I3Ipc::eventSocketStateChanged);
-	QObject::connect(&this->liveEventSocket, &QLocalSocket::readyRead, this, &I3Ipc::eventSocketReady);
-	QObject::connect(&this->liveEventSocket, &QLocalSocket::connected, this, &I3Ipc::subscribe);
+    QObject::connect(&this->liveEventSocket, &QLocalSocket::errorOccurred, this, &I3Ipc::eventSocketError);
+    QObject::connect(&this->liveEventSocket, &QLocalSocket::stateChanged, this, &I3Ipc::eventSocketStateChanged);
+    QObject::connect(&this->liveEventSocket, &QLocalSocket::readyRead, this, &I3Ipc::eventSocketReady);
+    QObject::connect(&this->liveEventSocket, &QLocalSocket::connected, this, &I3Ipc::subscribe);
 	// clang-format on
```
outfoxxed, line 89: Broken formatting

src/x11/i3/ipc/connection.cpp
```diff
@@ -74,52 +81,67 @@ I3Ipc::I3Ipc() {
...
+	auto type = QByteArray(std::bit_cast<std::array<char, 4>>(cmd).data(), 4);
+	auto len = QByteArray(std::bit_cast<std::array<char, 4>>(payloadLength).data(), 4);
+
+	return MAGIC.data() + len + type + payload;
 }
 
 void I3Ipc::subscribe() {
-	auto payload = QByteArray(R"(["workspace","output"])");
+	auto jsonArray = QJsonArray::fromStringList(this->mEvents);
+	const QJsonDocument jsonDoc(jsonArray);
```
outfoxxed: No constructors in this form please. Use auto or inline jsonDoc

src/x11/i3/ipc/listener.cpp
```diff
@@ -0,0 +1,43 @@
...
+void I3IpcListener::setSubscriptions(QList<QString> subscriptions) {
+	if (this->mSubscriptions == subscriptions) return;
+	this->mSubscriptions = std::move(subscriptions);
+
+	emit this->subscriptionsChanged();
+	this->startListening();
+}
+
+void I3IpcListener::startListening() {
+	if (this->i3Ipc != nullptr || this->mSubscriptions.isEmpty()) return;
```
outfoxxed: This wont allow subscriptions to be changed during operation. Why not replace the I3Ipc object instead?

outfoxxed (changes requested):
Can you rebase on master and reformat as well? Some recent changes to clang-format.

outfoxxed:
Thanks!

PR #363, opened 2025-11-15
wayland/layershell: add ability to set empty input region

bbedward:
Adds the ability to set an empty input region on wayland surfaces, which is useful for things like Wallpaper backgrounds.

```js
PanelWindow {
  WlrLayershell.inputMode = WlrInputMode.None
}
```

Without this, [labwc](https://labwc.github.io/)'s own input region is unable to receive events, which prevents its integrated desktop/context menus from working as expected. There are probably other use cases for this as well.

I just made it an enum with `None` and `Full (current and default behavior)` since I think that's the mai use case, don't see much use case for quickshell being able to support anything beyond that in terms of input regions.

### DIFF
--- a/src/wayland/wlr_layershell/surface.cpp
+++ b/src/wayland/wlr_layershell/surface.cpp
@@ -11,12 +11,14 @@
 #include <private/qwaylandwindow_p.h>
 #include <qlogging.h>
 #include <qrect.h>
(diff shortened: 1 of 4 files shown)


outfoxxed:
This is already exposed as `mask` which can be set to an empty region.

PR #375, opened 2025-11-24
services/tray: Signal tooltip changes to qml

topisani:
Previously, the tooltipTitleChanged and tooltipDescriptionChanged were never emitted.


outfoxxed:
thanks

PR #377, opened 2025-11-25
core/window: add startSystemMove, startSystemResize, maximize, and fullscreen

bbedward:
Allows to trigger move and resize operations on QsWindow, useful for making a built-in CSD, etc.

Also expose maximize and fullscreen methods and bindables.

### DIFF
--- a/src/window/floatingwindow.cpp
+++ b/src/window/floatingwindow.cpp
@@ -1,10 +1,12 @@
 #include "floatingwindow.hpp"
 
+#include <qnamespace.h>
 #include <qobject.h>
 #include <qqmlengine.h>
 #include <qqmllist.h>
 #include <qtmetamacros.h>
 #include <qtypes.h>
+#include <qwindow.h>
 
 #include "proxywindow.hpp"
 #include "windowinterface.hpp"
@@ -55,6 +57,7 @@ FloatingWindowInterface::FloatingWindowInterface(QObject* parent)
 	QObject::connect(this->window, &ProxyFloatingWindow::titleChanged, this, &FloatingWindowInterface::titleChanged);
 	QObject::connect(this->window, &ProxyFloatingWindow::minimumSizeChanged, this, &FloatingWindowInterface::minimumSizeChanged);
 	QObject::connect(this->window, &ProxyFloatingWindow::maximumSizeChanged, this, &FloatingWindowInterface::maximumSizeChanged);
+	QObject::connect(this->window, &ProxyWindowBase::windowConnected, this, &FloatingWindowInterface::onWindowConnected);
 	// clang-format on
 }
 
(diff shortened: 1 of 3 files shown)


outfoxxed:
Why is this exposed on ProxyWindowAttached and not FloatingWindow? Its illogical for other window types

bbedward:
> Why is this exposed on ProxyWindowAttached and not FloatingWindow? Its illogical for other window types



> Why is this exposed on ProxyWindowAttached and not FloatingWindow? Its illogical for other window types

I shifted it all to FloatingWindowInterface

src/window/floatingwindow.cpp
```diff
@@ -66,3 +69,52 @@ void FloatingWindowInterface::onReload(QObject* oldInstance) {
...
+	}
+
+	if (fullscreen != this->mFullscreen) {
+		this->mFullscreen = fullscreen;
+		emit this->fullscreenChanged();
+	}
+}
+
+bool FloatingWindowInterface::maximized() const { return this->mMaximized; }
+bool FloatingWindowInterface::fullscreen() const { return this->mFullscreen; }
```
outfoxxed: We can store the maximized or fullscreen flags for application before its visible, but should reflect the actual window state when queried (as well as be notified when it changes)

src/window/floatingwindow.hpp
```diff
@@ -78,15 +83,39 @@ class FloatingWindowInterface: public WindowInterface {
...
+	Q_INVOKABLE [[nodiscard]] bool startSystemResize(Qt::Edges edges) const;
+
+	/// Show the window in normal (restored) state.
+	Q_INVOKABLE void showNormal() const;
+	/// Show the window maximized.
+	Q_INVOKABLE void showMaximized() const;
+	/// Show the window minimized.
+	Q_INVOKABLE void showMinimized() const;
+	/// Show the window fullscreen.
+	Q_INVOKABLE void showFullScreen() const;
```
outfoxxed: should be named maximize() minimize() fullscreen()

src/window/floatingwindow.cpp
```diff
@@ -66,3 +69,52 @@ void FloatingWindowInterface::onReload(QObject* oldInstance) {
...
+	if (maximized != this->mMaximized) {
+		this->mMaximized = maximized;
+		emit this->maximizedChanged();
+	}
+
+	if (fullscreen != this->mFullscreen) {
+		this->mFullscreen = fullscreen;
+		emit this->fullscreenChanged();
+	}
+}
```
outfoxxed: Appears we actually want windowStates() instead of visibility? That also gives us minimize

src/window/floatingwindow.cpp
bbedward: Made the getters query windowStates directly, onVisibilityChanged emits the change signal

src/window/floatingwindow.hpp
bbedward: renamed

src/window/floatingwindow.cpp
bbedward: Changed to windowStates and added minimize to this

src/window/floatingwindow.cpp
bbedward: Actually connected to windowStateChanged, not visibility*

outfoxxed:
Reviewed this again and it seemed like max/min/full would work better as properties. I've made those changes, and am planning to merge once CI passes.

PR #381, opened 2025-11-26
hyprland/ipc: fix parsing order of windowTitle and windowClass in openwindow handler

nemalex:
The `openwindow` event format is `WINDOWADDRESS,WORKSPACENAME,WINDOWCLASS,WINDOWTITLE`, but the handler was parsing `args.at(2)` as title and `args.at(3)` as class, which is reversed.

This caused windows to display their class name instead of their actual title when the `openwindow` event arrived after `windowtitlev2`, since `updateInitial` would overwrite the correct title with the class.

Link to official IPC format: https://wiki.hypr.land/IPC/#events-list


outfoxxed:
thanks

PR #421, opened 2025-12-13
feat: add opt-in logind LockedHint for WlSessionLock

saltnpepper97:
Implements opt-in logind LockedHint updates for WlSessionLock, replacing the custom org.quickshell.SessionLock DBus interface.

## Changes

- Removed the custom org.quickshell.SessionLock DBus adaptor wiring.
- Updates logind LockedHint when WlSessionLock's existing exposeDbus opt-in is enabled.
- Sets LockedHint=true after the compositor confirms the lock is secure.
- Sets LockedHint=false when the session unlocks.
- Treats logind/DBus failures as warnings so locking is not blocked.
- Uses async DBus calls with generation tracking to avoid stale lock/unlock updates.

## Rationale

Uses logind's existing session LockedHint instead of a Quickshell-specific DBus interface. This matches the behavior expected by consumers that already observe logind session state, and avoids introducing a new custom lock-state API.

## Testing

- just configure debug -DNO_PCH=ON -DBUILD_TESTING=ON
- just build
- just test

Manual test:

```sh
~/.config/quickshell/locked-hint-test/watch-locked-hint.sh
just run
```

Click Lock in the test bar and confirm LockedHint changes to yes, then unlock and confirm it returns to no.


saltnpepper97:
Hello,

Sorry for the messy commits. This was my first time working with Qt, and it’s been a long time since I last touched C++. The linter and I spent some time fighting each other along the way.

Thanks for your patience!

bbedward:
I think that unconditionally registering the screensaver interface has a lot more considerations than just lock screen - because this would make it so nothing else could own the screensaver interface (like real screensaver apps, idle monitors, compositors)

- niri at least does this, so it can respect screensaver inhibit (treated as an idle inhibit)
- Inhibit is important for say, firefox calls inhibit on the screensaver interface when playing media (you'd usually want to create an idle inhibitor in that case to prevent locking/sleep)

If this is implemented in quickshell I think it'd make more sense as something that can be conditionally registered, since it wont work all the time anyway (like on niri)

```qml
Screensaver {
  onInhibited
  enabled
}
```

idk something like that, that's just my thoughts (not outfoxxed)

saltnpepper97:
> I think that unconditionally registering the screensaver interface has a lot more considerations than just lock screen - because this would make it so nothing else could own the screensaver interface (like real screensaver apps, idle monitors, compositors)
> 
>     * niri at least does this, so it can respect screensaver inhibit (treated as an idle inhibit)
> 
>     * Inhibit is important for say, firefox calls inhibit on the screensaver interface when playing media (you'd usually want to create an idle inhibitor in that case to prevent locking/sleep)
> 
> 
> If this is implemented in quickshell I think it'd make more sense as something that can be conditionally registered, since it wont work all the time anyway (like on niri)
> 
> ```qml
> Screensaver {
>   onInhibited
>   enabled
> }
> ```
> 
> idk something like that, that's just my thoughts (not outfoxxed)

@bbedward Been hard at work. After further inspection you were correct, that was an oversight on my part so instead what I did was make a custom interface for quickshell 

org.quickshell.SessionLock to bind to that instead so it shuld be free of other traffic as it's now custom :)

Just testing it at the moment. then linter fighting. Say any way to install tidyfox on arch linux without having to downloads the tar and unzip and just release and cp myself??? so that when testing my changes and fmt and such might be streamlined?

saltnpepper97:
@bbedward DONE, now it's on a custom interface no need to worry about ScreenSaver :)

djspacewhale:
Are there plans to merge this PR? Some of us downstream are waiting for this interface to fix some bugs.

outfoxxed:
Sorry I've missed this. I'm not sure why this isn't better accomplished with the IPC system though.

saltnpepper97:
Hello @outfoxxed, what exactly do you mean? this is in order to help programs such as stasis, a wayland idle manager that aims to be lock screen centric to track the locked state of quickshell. I have tried `bustctl` to no avail. If there is another way that I am unaware of I am open to understanding how that would work :)

> Sorry I've missed this. I'm not sure why this isn't better accomplished with the IPC system though.

outfoxxed:
Ok that makes more sense, I thought this was for individual configs, which should just be able to use the IPC system to get lockscreen signals.

@bbedward's approach using a more standard interface but explicit user registration makes sense to me on the surface, otherwise any application that would want to integrate with Quickshell here would need to special case it and it still might not match the user intent.

Anything I've missed for why we would want a custom interface here?

As a side note I'm not entirely sure what an idle manager is intended to do, but we do also implement [IdleMonitor](https://quickshell.org/docs/master/types/Quickshell.Wayland/IdleMonitor/) and [IdleInhibitor](https://quickshell.org/docs/master/types/Quickshell.Wayland/IdleInhibitor/).

saltnpepper97:
Thanks! the goal here isn’t per-config automation (IPC covers that well), it’s a stable integration point for external processes (idle managers, status monitors, etc.) to observe whether Quickshell’s lock UI is active.

I originally tried org.freedesktop.ScreenSaver, but as @bbedward noted that interface has broader ecosystem implications (idle daemons/compositors own it for inhibit semantics, e.g. niri treating inhibit as idle inhibit). I don’t want Quickshell to unconditionally claim that name and potentially break inhibit handling for apps like Firefox.

The reason for a custom org.quickshell.SessionLock interface is: it’s minimal (state + signals), doesn’t conflict with existing owners, and lets external apps integrate without going through the Quickshell IPC system.

That said, I agree that “standard interface when possible” is useful. I’m happy to add an explicit, opt-in/conditional registration for org.freedesktop.ScreenSaver (e.g. a Screensaver { enabled } block), where Quickshell only registers it if the bus name is unowned; otherwise it logs and continues. Inhibit semantics would remain owned by whoever already provides it.

Does that sound aligned with how you’d like Quickshell to expose lock state?

saltnpepper97:
@outfoxxed does this update with opt in via a config option help with that?

saltnpepper97:
@outfoxxed @bbedward, Have you looked over the new changes?? Basically now it's opt-in via an option 

example usage

```
import Quickshell
import Quickshell.Wayland
import QtQuick

ShellRoot {
  WlSessionLock {
    id: lock
    exposeDbus: true

    WlSessionLockSurface {
      Rectangle {
        anchors.fill: parent
        color: "#111111"

        Text {
          anchors.centerIn: parent
          text: "LOCKED – click to unlock"
          color: "white"
          font.pixelSize: 28
        }

        MouseArea {
          anchors.fill: parent
          onClicked: lock.locked = false
        }
      }
    }

    Component.onCompleted: lock.locked = true
  }
}

```

saltnpepper97:
Hi, sorry to bother again. I’m just curious as to why this PR is taking so long. I can show you a video in action if you’d like, to demonstrate how it works.

I mean, it’s a very simple change that gives you a custom interface for people to listen into (opt-in), simply to receive information on whether we’re currently locked or not.

Not only is this super useful for idle managers like Stasis, but think of the functionality this unlocks for scripting as well.

I hope we can find a way forward here.

bokicoder:
Should set logind LockedHint like niri; no need for a separate D-Bus interface just for lock state.

- https://github.com/niri-wm/niri/pull/1763
- https://github.com/niri-wm/niri/pull/3105#issuecomment-3692150573

saltnpepper97:
does this work now @bokicoder @outfoxxed @bbedward

saltnpepper97:
> Should set logind LockedHint like niri; no need for a separate D-Bus interface just for lock state.
> 
> - https://github.com/niri-wm/niri/pull/1763
> - https://github.com/niri-wm/niri/pull/3105#issuecomment-3692150573

Is this good now?

saltnpepper97:
Has anyone looked this over yet?

saltnpepper97:
Hi, I’d like to check whether this PR is still wanted.

The current version now uses opt-in logind `LockedHint` for `WlSessionLock`, based on the feedback in this thread. I’m happy to make further changes, but I need a clear direction: is this approach acceptable, should it be redesigned, or should I close the PR?

This has been open for several months, so even a “not interested” answer would be helpful for downstream users and for my own planning.

Thanks.

PR #429, opened 2025-12-15
all: initial support for freebsd

bigsaltyfishes:
Add initial support for FreeBSD.


outfoxxed:
This is still marked as draft, is it ready for review?

bigsaltyfishes:
Sorry for my late reply, i finished the test just now, should work on freebsd now.
image

.github/workflows/build.yml
```diff
@@ -61,3 +61,42 @@ jobs:
...
+              polkit \
+              glib \
+              mesa-libs \
+              git \
+              bash \
+              ninja \
+              gitup
+
+          run: |
+            cmake -GNinja -B build -DCRASH_REPORTER=OFF -DUSE_JEMALLOC=OFF -DCMAKE_C_FLAGS="-I/usr/local/include -L/usr/local/lib" -DCMAKE_CXX_FLAGS="-I/usr/local/include -L/usr/local/lib"
```
outfoxxed: Does Jemalloc not work correctly on freebsd?

src/services/pam/subprocess.cpp
```diff
@@ -83,7 +87,11 @@ PamIpcExitCode PamSubprocess::exec(const char* configDir, const char* config, co
 	logIf(this->log) << "Starting pam session for user \"" << user << "\" with config \"" << config
 	                 << "\" in dir \"" << configDir << "\"" << std::endl;
 
+#ifdef __FreeBSD__
+	auto result = pam_start(config, user, &conv, &handle);
+#else
 	auto result = pam_start_confdir(config, user, &conv, configDir, &handle);
+#endif
```
outfoxxed, line 94: If pam can't be configured on bsd a note of that should be made on the configdir property https://github.com/quickshell-mirror/quickshell/blob/3918290c1bcd93ed81291844d9f1ed146672dbfc/src/services/pam/qml.hpp#L39

.github/workflows/build.yml
bigsaltyfishes: Jemalloc is the allocator used by FreeBSD's libc. So I disabled it to let it directly use libc's malloc.

outfoxxed:
Dropped the link dirs, which I am not opposed to re-adding if it is actually the convention. Usually those paths are configured via the system and I haven't seen other packages do that. Otherwise LGTM

outfoxxed:
CI setup also failed on merge, so I've currently excluded it.

PR #475, opened 2026-01-13
network: provide static passwords without agent

cpwrs:
Add a WifiNetwork.passwordIsStatic property and password parameter to WifiNetwork.connect().

### DIFF
--- a/src/network/nm/backend.cpp
+++ b/src/network/nm/backend.cpp
@@ -127,8 +127,8 @@ void NetworkManager::registerDevice(const QString& path) {
 						managed ? this->registerFrontendDevice(type, dev) : this->removeFrontendDevice(dev);
 					};
 					// clang-format off
-					QObject::connect(dev, &NMDevice::requestAddAndActivateConnection, this, &NetworkManager::addAndActivateConnection);
-					QObject::connect(dev, &NMDevice::requestActivateConnection, this, &NetworkManager::activateConnection);
+					QObject::connect(dev, &NMDevice::sendAddAndActivateConnection, this, &NetworkManager::addAndActivateConnection);
+					QObject::connect(dev, &NMDevice::sendActivateConnection, this, &NetworkManager::activateConnection);
 					QObject::connect(dev, &NMDevice::managedChanged, this, onManagedChanged);
 					// clang-format on
 
--- a/src/network/nm/wireless.cpp
+++ b/src/network/nm/wireless.cpp
@@ -382,26 +382,13 @@ void NMWirelessDevice::registerFrontendNetwork(NMWirelessNetwork* net) {
 	// Bind WifiNetwork to NMWirelessNetwork
 	auto translateSignal = [net]() { return net->signalStrength() / 100.0; };
 	auto translateState = [net]() { return net->state() == NMConnectionState::Activated; };
-	auto passwordIsStatic = [net]() {
-		switch (net->security()) {
-		case WifiSecurityType::StaticWep:
-		case WifiSecurityType::WpaPsk:
-		case WifiSecurityType::Wpa2Psk:
-		case WifiSecurityType::Sae: return true;
-		default: return false;
-		}
-	};
-	auto connect = [this, net](const QString& /*password*/) {
+	auto handleConnect = [this, net](const QString& password = QString()) {
 		if (net->referenceConnection()) {
-			emit this->requestActivateConnection(net->referenceConnection()->path(), this->path());
+			emit this->sendActivateConnection(net->referenceConnection()->path(), this->path());
 			return;
 		}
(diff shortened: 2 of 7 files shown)


cpwrs:
Editing the raw settings map like this is isn't great, I'll make it a nice typed structure when I write the PR exposing Settings to the API.

src/network/wifi.hpp
```diff
@@ -103,16 +103,19 @@ class WifiNetwork: public Network {
...
 public:
 	explicit WifiNetwork(QString ssid, QObject* parent = nullptr);
 
 	/// Attempt to connect to the wifi network.
+	/// Passwords should only be provided to unknown networks with static passwords.
 	///
 	/// > [!WARNING] Quickshell does not yet provide a NetworkManager authentication agent,
 	/// > meaning another agent will need to be active to enter passwords for unsaved networks.
-	Q_INVOKABLE void connect();
+	Q_INVOKABLE void connect(const QString& password = QString());
```
cpwrs: Optional parameter or connectWithPassword?

src/network/nm/backend.cpp
```diff
@@ -241,10 +242,14 @@ void NetworkManager::activateConnection(
...
-    const QDBusObjectPath& specificObjectPath
+    const QString& devPath,
+    const QString& specificObjectPath
 ) {
-	auto pending = this->proxy->AddAndActivateConnection(settings, devPath, specificObjectPath);
+	auto pending = this->proxy->AddAndActivateConnection(
+	    settings,
+	    QDBusObjectPath(devPath),
+	    QDBusObjectPath(specificObjectPath)
+	);
```
outfoxxed, line 252: not sure why you're making these changes but I don't care either way

src/network/nm/wireless.cpp
```diff
@@ -382,41 +382,45 @@ void NMWirelessDevice::registerFrontendNetwork(NMWirelessNetwork* net) {
...
+	auto connect = [this, net](const QString& password) {
+		if (net->referenceConnection()) {
+			emit this->sendActivateConnection(net->referenceConnection()->path(), this->path());
+			return;
+		}
+		if (net->referenceAp()) {
+			this->addAndActivateConnection(net->referenceAp(), password);
+			return;
+		}
+	};
```
outfoxxed: Why not inline to QObject::connect?

src/network/nm/wireless.cpp
```diff
@@ -437,6 +441,55 @@ void NMWirelessDevice::removeNetwork() {
 	};
 }
 
+void NMWirelessDevice::addAndActivateConnection(NMAccessPoint* ap, const QString& password) {
+	auto settings = ConnectionSettingsMap();
+	auto passProvided = !password.isEmpty();
+
+	QVariantMap connection;
+	connection["autoconnect"] = true;
```
outfoxxed, line 440: Just noting that this would be very useful to have as an option on WifiNetwork in the future

src/network/wifi.hpp
outfoxxed: I think connectWithPassword would be clearer

outfoxxed (commented):
Looks fine, just a couple questions

src/network/nm/backend.cpp
cpwrs: Sticking with the pattern of only using the QDBusObjectPath when communicating with DBus, passing proxy->path() without having to cast it, etc.

src/network/wifi.hpp
```diff
@@ -122,22 +128,26 @@ class WifiNetwork: public Network {
...
 	void nmReasonChanged();
+	void passwordIsStaticChanged();
 
 private:
 	// clang-format off
 	Q_OBJECT_BINDABLE_PROPERTY(WifiNetwork, qreal, bSignalStrength, &WifiNetwork::signalStrengthChanged);
 	Q_OBJECT_BINDABLE_PROPERTY(WifiNetwork, bool, bKnown, &WifiNetwork::knownChanged);
 	Q_OBJECT_BINDABLE_PROPERTY(WifiNetwork, NMConnectionStateReason::Enum, bNmReason, &WifiNetwork::nmReasonChanged);
 	Q_OBJECT_BINDABLE_PROPERTY(WifiNetwork, WifiSecurityType::Enum, bSecurity, &WifiNetwork::securityChanged);
+	Q_OBJECT_BINDABLE_PROPERTY(WifiNetwork, bool, bPasswordIsStatic, &WifiNetwork::passwordIsStaticChanged);
```
outfoxxed, line 150: Didn't we determine that all passwords could be entered statically? I don't see this set anywhere

cpwrs:
I'll have a new one up this week that'll do all passwords

Issue #170, opened 2025-08-09
[Feature Request] add global menu


lisuke:
I want a global menu that handles all application menus with it.
like macos, KDE global menu, and so on.

outfoxxed:
Global menu is unfortunately a huge can of worms involving making custom qt platform interfaces, private gtk APIs etc. I'd like to see it eventually but its not planned for any time soon.

beta2ksu:
I wait in hope for this feature.

alicealysia:
For those interested, hellosystem has a page within their github dedicated to the subject: https://hellosystem.github.io/docs/developer/menu.html

As well as a working prototype for a global menu in wayland: https://github.com/helloSystem/Menu

In theory adding this to quickshell is a "simple" matter of compiling it into a qml module. That said, as mentioned by the developers themselves there's a *lot* of limitations and drawbacks.

This thread here contains a bunch of great info on  the subject and its challenges:

https://github.com/Alexays/Waybar/issues/288

PR #484, opened 2026-01-17
feat: expose DbusMenu to QML

CertainLach:
Can be used to implement global menus https://github.com/quickshell-mirror/quickshell/issues/170

Not sure if the appmenu component should be exposed by itself as it is done with status icons, given that it might be styled differently etc. Myself I have implemented global menus as a plugin for noctalia shell, and will try to publish it after I make its code somewhat readable

image

### DIFF
--- a/src/dbus/dbusmenu/dbusmenu.hpp
+++ b/src/dbus/dbusmenu/dbusmenu.hpp
@@ -171,10 +171,13 @@ class DBusMenuHandle;
 QDebug operator<<(QDebug debug, const DBusMenuHandle* handle);
 
 class DBusMenuHandle: public menu::QsMenuHandle {
-public:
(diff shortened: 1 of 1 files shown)


CertainLach:
Noctalia-shell plugin: https://github.com/CertainLach/noctalia-global-menu

outfoxxed (approved):
I can merge this as it might be more useful generally, but how are you piping in the global menus under wayland? Iirc there was no standard to be able to access them.

CertainLach:
> Iirc there was no standard to be able to access them.

I have implemented it using KDE protocol, which is the same canonical appmenu registrar, but the registration itself is performed using wayland protocol instead of dbus

Wayland protocol is implemented here: https://github.com/Naxdy/niri/pull/46

My implementation works for all QT and KDE apps

ryzendew:
After Testing this PR it works very well nice work!

PR #485, opened 2026-01-17
core: fix inverted inHeader condition in preprocessor

bbedward:
inHeader starts as false after latest commit, which breaks module loading. It should start as true and flip to off after it encounters the first {.


outfoxxed:
Thanks for the catch

Issue #486, opened 2026-01-17
pipewire reconnect breaks nixos-asahi


marcusramberg:
When running on https://github.com/quickshell-mirror/quickshell/commit/5d8354a88be2ce2c16add7457c94e29f6e7c3684 or newer on apple silicon (confirmed on m1 studio and macbook air) quickshell constantly scrolls messages like this:
```
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 7 with code -2 no global 60
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 0 with code -2 unknown resource 7 op:7
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 7 with code -2 no global 60
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 0 with code -2 unknown resource 7 op:7
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 7 with code -2 no global 60
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 0 with code -2 unknown resource 7 op:7
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 7 with code -2 no global 60
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 0 with code -2 unknown resource 7 op:7
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 7 with code -2 no global 60
  WARN quickshell.service.pipewire.loop: Fatal pipewire error on object 0 with code -2 unknown resource 7 op:7
```

And DMS sound overlay and sound selection goes kinda crazy with changing between 'No output' and the actual one.

PR #490, opened 2026-01-18
services/pipewire: make ENOENT non-fatal

marcusramberg:
Fixes https://github.com/quickshell-mirror/quickshell/issues/486 as it only logs a warning on startup. DMS input/output selector works as normal.

### DIFF
--- a/src/services/pipewire/core.cpp
+++ b/src/services/pipewire/core.cpp
@@ -143,6 +143,15 @@ void PwCore::onSync(void* data, quint32 id, qint32 seq) {
 void PwCore::onError(void* data, quint32 id, qint32 /*seq*/, qint32 res, const char* message) {
 	auto* self = static_cast<PwCore*>(data);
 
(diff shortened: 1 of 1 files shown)


outfoxxed:
Is there any determination for why we're excluding enoent in particular apart from it being the one that was failing? Pipewire error codes seemed to just come from the underlying syscalls instead of being distinct for any particular type of error.

marcusramberg:
My thinking was that ENOENT seemed to indicate some missing device and probably wasn't fatal, but I don't have any deep insights in pipewire so I'm totally willing to accept it might not be the right fix.

outfoxxed:
It seems like you're right and maybe I misremembered something. Spent a couple hours digging through it and testing behavior and ended up at the same conclusion of just ignoring ENOENT, and merged something equivalent.

It appears pipewire's error codes come from both itself AND syscalls it makes, and it just tries to pick ones that kind of match, which honestly seems insane.

PR #498, opened 2026-01-20
core/proxywindow: create window on visibility for lazily initialized windows

bbedward:
de1bfe028d6982ac9dce08e5063ea5611498b204 introduced a regression for windows where they would not be created if created initially with visible:false. This is a fix for that, by creating them if window is null but visible is true.


outfoxxed:
Thanks for the fix

PR #500, opened 2026-01-21
network: add freebsd backend

charlesrocket:
This adds a native FreeBSD backend for the network module. Tested with `network.qml`/[widget](https://github.com/charlesrocket/dotfiles/blob/ec03b44ead9cd4bc389343f675348a2dfb42145e/.config/quickshell/components/Network.qml).

### DIFF
--- a/src/network/CMakeLists.txt
+++ b/src/network/CMakeLists.txt
@@ -1,4 +1,8 @@
-add_subdirectory(nm)
+if(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
+	add_subdirectory(freebsd)
+else()
+	add_subdirectory(nm)
+endif()
 
 qt_add_library(quickshell-network STATIC
 	network.cpp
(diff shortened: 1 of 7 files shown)


cpwrs:
Is there a reason to use wpa_cli instead of the control interface directly?
https://w1.fi/wpa_supplicant/devel/ctrl_iface_page.html

wpa_supplicant is a valid backend for linux as well. Is there other networking stacks for FreeBSD? If so, this should probably be the FreeBSD impl of the wpa_supplicant backend and not the other way around.

I was planning on adding IWD as a backend eventually, which is a similar problem. I think that these backends should cover only the portion of the API they can support alone. For example, the Linux IWD backend wouldn't use udev to list non-wireless devices.

charlesrocket:
I used CLI because I am familiar with it. Right now I just want the current FBSD PRs merged, then I can start improving these modules.

charlesrocket:
After the second and obviously deliberate line cut, I am done with this circus.

PR #532, opened 2026-02-01
core: improve named icon detection

charlesrocket:
This fixes weak named icon detection. The provider now can look for light or dark icons depending on the system settings (KDE/GTK).

Valid paths (that were undetectable before):

```
/usr/local/share/icons/breeze/status/24/transmission-tray-icon.svg
/usr/local/share/icons/breeze-dark/status/22/transmission-tray-icon.svg
```

Now the provider will pull the icon from `breeze-dark` if dark icons are preferred.

Tested on FreeBSD via `transmission-gtk`/`transmission-qt`.

### DIFF
--- a/src/core/iconimageprovider.cpp
+++ b/src/core/iconimageprovider.cpp
@@ -2,18 +2,135 @@
 #include <algorithm>
 
 #include <qcolor.h>
+#include <qcontainerfwd.h>
+#include <qdir.h>
+#include <qfile.h>
 #include <qicon.h>
 #include <qlogging.h>
 #include <qpainter.h>
 #include <qpixmap.h>
+#include <qsettings.h>
 #include <qsize.h>
 #include <qstring.h>
+
+namespace {
+// Get the system's color scheme preference
+bool getSystemDarkPreference() {
+	// Check KDE
+	const QSettings kdeSettings(QDir::homePath() + "/.config/kdeglobals", QSettings::IniFormat);
+	if (kdeSettings.value("General/ColorScheme").toString().toLower().contains("dark")) {
+		return true;
+	}
+
+	// Check GTK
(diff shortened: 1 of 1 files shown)


outfoxxed:
This is generally supposed to be handled by qt, and the qpa selecting light or dark icons.
We do eventually want to rewrite this, but a more complicated rewrite is required here. Notably it needs to be async and not hardcode anything but fallback paths.

charlesrocket:
Pulling the GTK preference would probably require a hardcoded path to config, not sure if I can reach it via Qt API.

outfoxxed:
This shouldnt require checking the gtk theme. If the user configured Qt properly, then the information should be available from the QPA.

charlesrocket:
You mean GNOME settings to KDE settings configuration?

outfoxxed:
No. Most features of Quickshell don't even work under GNOME since they don't implement the protocols necessary for it, and I'm not aware of anyone using it under KDE, as plasmashell has much better insight into the DE.

I'm talking about standalone QPA themes such as [qtengine](https://github.com/kossLAN/qtengine) and qt(5/6)ct, or KDE's QPA, which are supposed to back Qt's icon theme loading, including color selection. For example my icon theme is "breeze-dark" set by qtengine.

charlesrocket:
Oh I see, I'll check it out.

outfoxxed:
Closing this under the assumption you were missing QPA setup. Reopen if still needed.

charlesrocket:
```
QT_QPA_PLATFORMTHEME=qt6ct
QT_QPA_PLATFORM=wayland;xcb
```

charlesrocket:
Tried to set via pragma too without positive results, so this is the only way I cold get icons to work (maybe keep this open).

PR #535, opened 2026-02-02
network: add NMSettings and connectWithPsk/Settings

cpwrs:
This exposes the connection settings from NetworkManager.

Additions:
- NMSettings class accessible via Network.nmSettings
- Network.connectionFailed(ConnectionFailReason) signal
- Network.connectWithPsk()
- Network.connectWithSettings() for the multi-setting network use case
- Reference types where possible in doc comments

Fixes:
- Unmanaged devices are now included in the model, and NetworkDevice.nmManaged is a writable prop
- Moved WifiNetwork.{connect, disconnect, forget}() to its parent class Network
- securityFromConnectionSettings correctly handles Unknown, Open, and Owe


outfoxxed:
noticing you formatted it, is this ready now?

cpwrs:
> noticing you formatted it, is this ready now?


Not yet, gotta add back some password stuff. I have time to work on it today

src/network/test/manual/network.qml
```diff
@@ -85,17 +96,50 @@ FloatingWindow {
...
 						Layout.fillWidth: true
-						model: {
-							if (modelData.type !== DeviceType.Wifi) return []
-							return [...modelData.networks.values].sort((a, b) => {
-								if (a.connected !== b.connected) {
-									return b.connected - a.connected
-								}
-								return b.signalStrength - a.signalStrength
-							})
-						}
```
cpwrs, line 96: This makes the model re eval and resets the state. Didn't notice that till I added a password prompt.

src/network/network.cpp
```diff
@@ -62,4 +55,73 @@ Network::Network(QString name, QObject* parent): QObject(parent), mName(std::mov
...
+			if (failReason == NMDeviceStateReason::NoSecrets) emit this->noSecrets();
+			if (failReason == NMDeviceStateReason::SupplicantDisconnect)
+				emit this->supplicantDisconnect();
+			if (failReason == NMDeviceStateReason::SupplicantFailed) emit this->supplicantFailed();
+			if (failReason == NMDeviceStateReason::SupplicantTimeout) emit this->supplicantTimeout();
+		};
+	});
+
+	connect(network, &Network::destroyed, this, [this]() { this->bNetwork = nullptr; });
+}
```
cpwrs: I included a really basic set, but it could include signals for anything from
https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMDeviceStateReason
or
https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMActiveConnectionStateReason

The latter unfortunately only emits for VPN connections, for wifi conns its always DeviceDisconnected. 

Sadly there's no "auth failed", that falls under NoSecrets

src/network/network.hpp
```diff
@@ -105,38 +89,132 @@ private slots:
...
+/// It emits useful signals about attempts to connect to the network.
+class NMConnectionContext: public QObject {
+	Q_OBJECT;
+	QML_ELEMENT;
+
+	// clang-format off
+	/// A network to provide connection context for or `null`.
+	Q_PROPERTY(Network* network READ network WRITE setNetwork NOTIFY networkChanged)
+	/// The last connection settings that context signals were emitted for or `null`.
+	Q_PROPERTY(NMSettings* settings READ default NOTIFY settingsChanged BINDABLE bindableSettings)
```
cpwrs: Is this what you were thinking for a "callback" to the specific settings?

src/network/nm_settings.cpp
```diff
@@ -0,0 +1,76 @@
...
+	if (psklen == 64) {
+		for (int i = 0; i < psklen; ++i) {
+			if (!psk.at(i).isLetterOrNumber()) {
+				return false;
+			}
+		}
+	}
+
+	return true;
+}
```
cpwrs: Should I be validating these?

cpwrs:
I couldn't find a useful way to expose the settings as a big QMap (or multiple smaller maps), so it's empty besides the psk setter. Adding invokables for the popular eap methods is easy enough but I wasn't able to test them.

As far as connection context, I was disappointed to learn that only VPN connections hit the more descriptive state reasons in https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMActiveConnectionStateReason. 
I was also not sure how the context should be implemented in QML. Let me know how I can clean this up!

src/network/test/manual/network.qml
outfoxxed: ScriptModel can be used to internally diff and not reset

outfoxxed:
Just checking in on this because it'd be nice to have in 0.3 which I want to cut this month. I should be around to review on more sane timelines than usual.

src/network/nm/dbus_types.hpp
```diff
@@ -1,9 +1,68 @@
...
+	quint32 metric = 0;
+};
+
+const QDBusArgument& operator>>(const QDBusArgument& argument, qs::network::NMIPv6Route& route);
+const QDBusArgument& operator<<(QDBusArgument& argument, const qs::network::NMIPv6Route& route);
+
+const QDBusArgument&
+operator>>(const QDBusArgument& argument, QList<qs::network::NMIPv6Route>& list);
+const QDBusArgument&
+operator<<(QDBusArgument& argument, const QList<qs::network::NMIPv6Route>& list);
```
cpwrs: These struct settings are deprecated on stable but are still initialized by NM by default

src/network/nm/settings.hpp
```diff
@@ -0,0 +1,71 @@
...
+	/// Only changed fields need to be included.
+	///
+	/// > [!NOTE] Secrets may be part of the update request,
+	/// > and will be either stored in persistent storage or sent to a Secret Agent for storage,
+	/// > depending on the flags associated with each secret.
+	Q_INVOKABLE void write(const QVariantMap& settings);
+	/// Get the settings map describing this network configuration.
+	///
+	/// > [!NOTE] This will never include any secrets required for connection to the network, as those are often protected.
+	Q_INVOKABLE QVariantMap read();
```
cpwrs, line 50: This can just be a Q_PROPERTY now if that's preferred. I can throw in read only props for id/uuid too for convenience.

write() is nice because you only have to give it the fields you want changed. Right now this provides setting = null to remove/unset a setting.

src/network/nm/dbus_types.hpp
```diff
@@ -1,9 +1,63 @@
...
+} // namespace qs::network
+
+const QDBusArgument& operator>>(const QDBusArgument& argument, QList<QVariantMap>& list);
+const QDBusArgument& operator<<(QDBusArgument& argument, const QList<QVariantMap>& list);
+const QDBusArgument& operator>>(const QDBusArgument& argument, QList<quint32>& list);
+const QDBusArgument& operator<<(QDBusArgument& argument, const QList<quint32>& list);
+const QDBusArgument& operator>>(const QDBusArgument& argument, QList<QList<quint32>>& list);
+const QDBusArgument& operator<<(QDBusArgument& argument, const QList<QList<quint32>>& list);
+const QDBusArgument& operator>>(const QDBusArgument& argument, QList<QByteArray>& list);
+const QDBusArgument& operator<<(QDBusArgument& argument, const QList<QByteArray>& list);
```
cpwrs: Didn't type alias these because more than one setting uses these signatures and this would be the correct operator for them in any context. 

I can give them some general aliases and bring them into network scope if you prefer

src/network/network.hpp
```diff
@@ -105,38 +89,132 @@ private slots:
...
+	QML_UNCREATABLE("Network can only be aqcuired through networking devices");
 
 	// clang-format off
 	/// The name of the network.
 	Q_PROPERTY(QString name READ name CONSTANT);
+	/// A list of connnection settings profiles for this network.
+	///
+	/// > [!WARNING] Only valid for the NetworkManager backend. 
+	QSDOC_TYPE_OVERRIDE(ObjectModel<qs::network::NMSettings>*);
+	Q_PROPERTY(UntypedObjectModel* nmSettings READ nmSettings CONSTANT);
```
outfoxxed: QList please

src/network/device.hpp
```diff
@@ -85,9 +27,19 @@ class NetworkDevice: public QObject {
...
+	Q_PROPERTY(NMDeviceState::Enum nmState READ default NOTIFY nmStateChanged BINDABLE bindableNmState);
+	/// A reason for the @@nmState of the device.
+	///
+	/// > [!WARNING] Only valid for the NetworkManager backend.
+	Q_PROPERTY(NMDeviceStateReason::Enum nmStateReason READ default NOTIFY nmStateReasonChanged BINDABLE bindableNmStateReason);
+	/// True if the device is managed by NetworkManager.
+	///
+	/// > [!WARNING] Only valid for the NetworkManager backend.
+	Q_PROPERTY(bool nmManaged READ nmManaged WRITE setNmManaged NOTIFY nmManagedChanged)
+	/// True if the device is allowed to autoconnect to a network.
```
outfoxxed, line 38: As we've discussed these will need to move once we work more on alternate backends. I can accept them for now but we don't want them on the base objects.

src/network/network.cpp
```diff
@@ -55,11 +51,85 @@ void Networking::setWifiEnabled(bool enabled) {
...
+		if (network->nmStateReason() == NMNetworkStateReason::DeviceDisconnected) {
+			auto failReason = network->nmDeviceFailReason();
+			if (failReason == NMDeviceStateReason::NoSecrets) emit this->noSecrets();
+			if (failReason == NMDeviceStateReason::SupplicantDisconnect)
+				emit this->supplicantDisconnect();
+			if (failReason == NMDeviceStateReason::SupplicantFailed) emit this->supplicantFailed();
+			if (failReason == NMDeviceStateReason::SupplicantTimeout) emit this->supplicantTimeout();
+			if (failReason == NMDeviceStateReason::SsidNotFound) emit this->wifiNetworkLost();
+		};
+	});
```
outfoxxed: Again direct nm usage in a base class.

src/network/network.hpp
```diff
@@ -101,42 +71,177 @@ private slots:
...
+	void supplicantTimeout();
+	/// The WiFi network could not be found
+	void wifiNetworkLost();
+
+private:
+	// clang-format off
+	Q_OBJECT_BINDABLE_PROPERTY(ConnectionContext, Network*, bNetwork, &ConnectionContext::networkChanged);
+	Q_OBJECT_BINDABLE_PROPERTY(ConnectionContext, NMSettings*, bSettings, &ConnectionContext::settingsChanged);
+	// clang-format on
 };
```
outfoxxed: I realize i suggested this before but I think the connection context isnt super helpful here especially with all the NM ties and it not managing lifecycle. I'm thinking remove it for now, we can potentially add it back later.

src/network/nm/wireless.cpp
```diff
@@ -385,46 +390,118 @@ void NMWirelessDevice::registerFrontendNetwork(NMWirelessNetwork* net) {
...
+			    QObject::connect(
+			        call,
+			        &QDBusPendingCallWatcher::finished,
+			        this,
+			        [this, ref](QDBusPendingCallWatcher* call) {
+				        const QDBusPendingReply<> reply = *call;
+
+				        if (reply.isError()) {
+					        qCWarning(logNetworkManager) << "Failed to write PSK for" << this->path() << ":"
+					                                     << reply.error().message();
```
outfoxxed: These types of failures should be INFO over WARN because WARN is always shown and we generally don't like to pollute default visibility logs

src/network/wifi.cpp
```diff
@@ -6,99 +6,35 @@
...
 	}
-
-	this->requestConnect();
-}
-
-void WifiNetwork::disconnect() {
-	if (!this->bConnected) {
-		qCCritical(logWifi) << this << "is not currently connected";
+	if (this->bSecurity != WifiSecurityType::WpaPsk && this->bSecurity != WifiSecurityType::Wpa2Psk
+	    && this->bSecurity != WifiSecurityType::Sae)
```
outfoxxed, line 30: Unsure if its the same thing but it looks like there's another potentially psk requiring security type

src/network/nm/utils.cpp
```diff
@@ -1,27 +1,28 @@
 #include "utils.hpp"
-
+#include <cmath>
 // We depend on non-std Linux extensions that ctime doesn't put in the global namespace
 // NOLINTNEXTLINE(modernize-deprecated-headers)
 #include <time.h>
```
outfoxxed: Where is this happening and are these extensions BSD compatible? We have contributors actively trying for BSD compatibility.

src/network/nm/dbus_types.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+const QDBusArgument& operator<<(QDBusArgument& argument, const QList<QByteArray>& list) {
+	argument.beginArray(qMetaTypeId<QByteArray>());
+	for (const auto& val: list) {
+		argument << val;
+	}
+	argument.endArray();
+	return argument;
+}
```
outfoxxed: This particular abomination appears to be entirely solvable with templates when testing locally. (qdebugs showed they were called)

```cpp
template <typename T>
const QDBusArgument& operator>>(const QDBusArgument& argument, QList<T>& list) {
	argument.beginArray();
	while (!argument.atEnd()) list.append(qdbus_cast<T>(argument));
	argument.endArray();
	return argument;
}

template <typename T>
const QDBusArgument& operator<<(QDBusArgument& argument, const QList<T>& list) {
	argument.beginArray(qMetaTypeId<T>());
	for (const auto& v: list) argument << v;
	argument.endArray();
	return argument;
}
```

It looks like something similar is defined in qt as well, but not sure if we can use it or we're supposed to redefine it.

src/network/wifi.cpp
cpwrs: I can include the standard StaticWep case here. Anything else requires arbitrary EAP secrets to be written.
WEP requires specifying a key index, but it's almost always index 0.

Thought about removing WEP support altogether because it's ancient and useless. It was never supported by IWD, and disabled in default wpa_supplicant builds:
https://git.w1.fi/cgit/hostap/tree/wpa_supplicant/defconfig

src/network/nm/wireless.cpp
cpwrs: What constitutes a WARN? I moved any warnings that could come from user input to INFO.

src/network/nm/dbus_types.cpp
cpwrs: Seems like using the qt definitions is working

src/network/device.hpp
cpwrs: Removed NMDeviceState and NMDeviceStateReason. 

Reasons are usually only needed after a failure so that's handled in the connectionFailed signal I added.

nmManaged and nmSettings is NM-specific, not sure how you want to handle those. I could see the argument that choosing managed devices shouldn't be done within the client.

src/network/enums.hpp
```diff
@@ -21,189 +21,47 @@ class DeviceConnectionState: public QObject {
...
+		/// Secrets were required, but not provided.
+		NoSecrets = 1,
+		/// The 802.1x supplicant disconnected.
+		SupplicantDisconnected = 2,
+		/// The 802.1x supplicant failed.
+		SupplicantFailed = 3,
+		/// The 802.1x supplicant took too long to authenticate.
+		SupplicantTimeout = 4,
+		/// The WiFi network could not be found.
+		WifiNetworkLost = 5,
```
outfoxxed, line 85: Can you elaborate a bit in the docstrings / rename them for clarity? I have a feeling 90% of users will not know what an 802.1x supplicant is.

src/network/nm/settings.cpp
```diff
@@ -117,8 +117,8 @@ void NMSettings::clearSecrets() {
 		const QDBusPendingReply<> reply = *call;
 
 		if (reply.isError()) {
-			qCWarning(logNetworkManager)
-			    << "Failed to clear secrets for" << this->path() << ":" << reply.error().message();
+			qCInfo(logNetworkManager) << "Failed to clear secrets for" << this->path() << ":"
+			                          << reply.error().message();
 		} else {
```
outfoxxed, line 122: Distinction here and for the lower warn changes: These should stay warn, as they're issues in qs / nm / not a user mistake. A bad password doesn't need to warn, but it makes sense when something is actually misbehaving.

PR #544, opened 2026-02-06
core: add hasIconTheme mapping

kossLAN:
Adds a mapping to [QIcon::hasIconTheme](https://doc.qt.io/qt-6/qicon.html#hasThemeIcon), useful for checking if an icon provided by systray or similar scenarios is from your icon theme or not, as far as I know there is no way of checking this via quickshell ATM.


kossLAN:
messed up name in commit messages woops, its hasThemeIcon

kossLAN:
*nothing happened* i love git

PR #552, opened 2026-02-09
nix: use libxcb directly

Moraxyc:
```
evaluation warning: The xorg package set has been deprecated, 'xorg.libxcb' has been renamed to 'libxcb'
```
In https://github.com/NixOS/nixpkgs/pull/479724, package set `xorg` has been deprecated.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -18,7 +18,7 @@
   wayland,
   wayland-protocols,
   wayland-scanner,
-  xorg,
+  libxcb,
   libdrm,
   libgbm ? null,
   vulkan-headers,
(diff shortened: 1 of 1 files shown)


outfoxxed:
This can be merged once a stable nix release includes that commit

musjj:
`libxcb` has always been available as a top-level package since at least `25.11` (stable): https://search.nixos.org/packages?channel=25.11&query=libxcb

So I don't think this change would break any stable user.

LuckShiba:
I could build this PR using nixpkgs 25.11 with no errors or warnings, it should be fine for all 25.11 users.

PR #566, opened 2026-02-13
wayland/background-effect: add ext-background-effect-v1 (blur region)

bbedward:
core/region: add radius to create rounded regions

Adds `BackgroundEffect` that allows setting blur region on layer shell surfaces, additionally adds a radius option to `Region` similar to the vicinae pattern for making approximate rounded regions.

To test on a supported compositor: `qs -p src/wayland/background_effect/test/manual/background_effect.qml`

Usage is as:

```qml
BackgroundEffect.blurRegion: Region {
        item: root.contentItem
}
```

### DIFF
--- a/src/core/region.cpp
+++ b/src/core/region.cpp
@@ -1,4 +1,5 @@
 #include "region.hpp"
+#include <algorithm>
 #include <cmath>
 
 #include <qobject.h>
@@ -18,6 +19,11 @@ PendingRegion::PendingRegion(QObject* parent): QObject(parent) {
 	QObject::connect(this, &PendingRegion::yChanged, this, &PendingRegion::changed);
 	QObject::connect(this, &PendingRegion::widthChanged, this, &PendingRegion::changed);
 	QObject::connect(this, &PendingRegion::heightChanged, this, &PendingRegion::changed);
+	QObject::connect(this, &PendingRegion::radiusChanged, this, &PendingRegion::changed);
+	QObject::connect(this, &PendingRegion::topLeftRadiusChanged, this, &PendingRegion::changed);
+	QObject::connect(this, &PendingRegion::topRightRadiusChanged, this, &PendingRegion::changed);
+	QObject::connect(this, &PendingRegion::bottomLeftRadiusChanged, this, &PendingRegion::changed);
+	QObject::connect(this, &PendingRegion::bottomRightRadiusChanged, this, &PendingRegion::changed);
 	QObject::connect(this, &PendingRegion::childrenChanged, this, &PendingRegion::changed);
 }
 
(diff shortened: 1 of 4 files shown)


YaLTeR:
Seems the region passed to the protocol also needs to be offset by the Qt CSD, if any:

image

Better visibility of the region:

image

outfoxxed:
For consistency with existing interfaces, this should probably be an attached object. See HyprlandWindow for an example. 

Do we have any options for blur opacity exposed?

bbedward:
> For consistency with existing interfaces, this should probably be an attached object. See HyprlandWindow for an example.

Changed it to an attached object, the only downside I guess is my thought was getting a BackgroundEffect.active to tell if the compositor had applied it/supported it
 
> Do we have any options for blur opacity exposed?

That's not exposed by the protocol, its just `blurRegion` and `capabilities { blur }` pretty much. The effect parameters themselves would be set elsewhere, like niri's compositor config in this case.

src/core/region.cpp
```diff
@@ -90,6 +92,25 @@ QRegion PendingRegion::build() const {
...
+			auto dx = static_cast<double>(r * r - (r - i) * (r - i));
+			auto cut = r - static_cast<int>(std::round(std::sqrt(dx)));
+			if (cut <= 0) continue;
+			region -= QRegion(x, y + i, cut, 1);
+			region -= QRegion(x + w - cut, y + i, cut, 1);
+			region -= QRegion(x, y + h - 1 - i, cut, 1);
+			region -= QRegion(x + w - cut, y + h - 1 - i, cut, 1);
+		}
+	}
+
```
outfoxxed: I'm not quite sure about this one for two reasons:
1. Matching exactly what qt does for rounded rectangle corners is a little iffy
2. Corners should be an easier primitive to apply than this, especially when you don't need exactly four square ones.

The best way I see to solve both problems at once is to use one QRegion(Ellipse) per corner and two QRegion(Rectangle)s forming a + between the ellipse centers, then unlock the corners.

In addition to potentially matching qt's rects a little more closely, unlocking corners (and removing the /2 constraint) lets users build corner pieces which can be intersected with other shapes when the blurred area is a little more custom.

src/wayland/background_effect/surface.cpp
```diff
@@ -0,0 +1,44 @@
...
+    : QtWayland::ext_background_effect_surface_v1(surface) {}
+
+BackgroundEffectSurface::~BackgroundEffectSurface() {
+	if (!this->isInitialized()) return;
+
+	if (this->mInert) {
+		wl_proxy_destroy(reinterpret_cast<wl_proxy*>(this->object())); // NOLINT(misc-include-cleaner)
+	} else {
+		this->destroy();
+	}
```
outfoxxed: Protocol objects should always use their actual destructors. That wl_proxy_destroy is doing at best the exact same thing or at worst leaking a protocol object id.

src/wayland/background_effect/surface.cpp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void BackgroundEffectSurface::setInert() { this->mInert = true; }
```
outfoxxed: Inert is a useless state as its not clearly defined if we can ever use the object again, so we have to assume we can't. In any case where the effect would become inert it should be destroyed.

src/wayland/background_effect/qml.cpp
```diff
@@ -0,0 +1,221 @@
...
+	if (!this->surface || !this->pendingBlurRegion) return;
+
+	QRegion region;
+	if (this->mBlurRegion) {
+		region =
+		    this->mBlurRegion->applyTo(QRect(0, 0, this->mWindow->width(), this->mWindow->height()));
+
+		auto margins = this->mWaylandWindow->clientSideMargins();
+		region.translate(margins.left(), margins.top());
+	}
```
outfoxxed, line 115: Scaling is not handled. The surface-local coordinates are the worst part of this generally terrible protocol in that they'll make highdpi look ugly and viewporter cant fix it (afaik), but Qt's internal scaling mechanisms like QT_SCALE_FACTOR have to be accounted for because some people use them for changing logical pixel size in qs.

src/core/region.cpp
bbedward: Replaced with the ellipse + cross approach and removed the /2 constraint

src/wayland/background_effect/surface.cpp
bbedward: Removed. The destructor now always calls destroy(). To ensure it runs while the surface is still alive, I added SurfaceAboutToBeDestroyed event filter, which fires before reset() tears down the surface.

src/wayland/background_effect/surface.cpp
bbedward: Removed inert

src/wayland/background_effect/qml.cpp
bbedward: Added QHighDpiScaling::factor(this->mWindow) to account for scaling

src/wayland/background_effect/qml.cpp
```diff
@@ -0,0 +1,246 @@
...
+		auto* surfaceEvent = dynamic_cast<QPlatformSurfaceEvent*>(event);
+		if (surfaceEvent->surfaceEventType() == QPlatformSurfaceEvent::SurfaceAboutToBeDestroyed) {
+			this->surface = nullptr;
+			this->pendingBlurRegion = false;
+		}
+	}
+
+	return this->QObject::eventFilter(object, event);
+}
+
```
outfoxxed, line 132: The intent here is good, but the cleanup isn't super time-sensitive, so onWaylandSurfaceDestroyed's = nullptr should cover it fine already, and pendingBlurRegion is set on surface creation.

src/core/region.cpp
```diff
@@ -18,6 +18,7 @@ PendingRegion::PendingRegion(QObject* parent): QObject(parent) {
 	QObject::connect(this, &PendingRegion::yChanged, this, &PendingRegion::changed);
 	QObject::connect(this, &PendingRegion::widthChanged, this, &PendingRegion::changed);
 	QObject::connect(this, &PendingRegion::heightChanged, this, &PendingRegion::changed);
+	QObject::connect(this, &PendingRegion::radiusChanged, this, &PendingRegion::changed);
 	QObject::connect(this, &PendingRegion::childrenChanged, this, &PendingRegion::changed);
```
outfoxxed, line 27: I don't want to expand scope on this much for a mostly unrelated PR but per-corner radius control would still be very useful here for things like inverse corners. Though you can make them without a builtin its much easier with one that gets you most of the way there.

src/wayland/background_effect/qml.cpp
bbedward: I couldn't get around this with this protocol at least specifically, surfaceDestroyed fires after wl_surface_destroy is queued - and the compositor raises a surface_destroyed error (crashes quickshell) I guess because the objects are inert and the spec of this protocol defines that as an error case - but others like idle-inhibit don't.

So the SurfaceAboutToBeDestroyed seems needed to make sure cleanup happens before wl_surface_destroy.

src/core/region.cpp
bbedward: Added per-corner radius control

src/wayland/background_effect/qml.cpp
outfoxxed: Reasonable solution I guess. Feels non ideal but I won't block the pr on it.

src/core/region.cpp
outfoxxed: image

Doesn't really work. At this point we could revert or fix the independent radius changes, your choice, since I don't want to block the main pr on what is essentially a side issue.

bbedward: I changed the approach, to a 5 rectangle approach center and each edge. That is working for me in the test cases I tried anyway

outfoxxed:
Added missing controls to the tester and noticed the corner radius is entirely unclamped at this point so it will break if you take it out of bounds. It should clamp at the point where its bounded by the size of the rect or space used by other corners

bbedward:
> Added missing controls to the tester and noticed the corner radius is entirely unclamped at this point so it will break if you take it out of bounds. It should clamp at the point where its bounded by the size of the rect or space used by other corners

Added the clamping to region

outfoxxed:
It seems like we're still getting intersections from circles in opposite corners

bbedward:
> It seems like we're still getting intersections from circles in opposite corners

I changed the approach again, to cut the corners out of the rectangle. All the scenarios I tried in the tester seem to work for me

outfoxxed:
Modified to use ext-bg-effect from w-p and split commits. Will merge once CI passes.

Issue #534, opened 2026-02-01
Crash when remote disconnects from Sunshine


mgabor3141:
### General crash information

<details> <summary>General information</summary>

```
===== Build Information =====
Git Revision: 1e4d804e7f3fa7465811030e8da2bf10d544426a
Buildtime Qt Version: 6.10.2
Build Type: RelWithDebInfo
Compiler: GNU (15.2.1)
Complie Flags: -march=native -O3 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -Wp,-D_GLIBCXX_ASSERTIONS -flto=auto

Build configuration:
  Distributor: AUR (package: quickshell-git)
  Distributor provided debuginfo: NO
  Disable precompild headers (dev): OFF
  Build tests (dev): OFF
  ASAN (dev): OFF
  Keep Frame Pointers (dev): OFF
  Crash Handling: ON
  Use jemalloc: ON
  Unix Sockets: ON
  Wayland: ON
    Wlroots Layer-Shell: ON
    Session Lock: ON
    Foreign Toplevel Management: ON
    Hyprland: ON
      Hyprland IPC: ON
      Hyprland Global Shortcuts: ON
      Hyprland Focus Grabbing: ON
      Hyprland Surface Extensions: ON
    Screencopy: ON
      Image Copy Capture: ON
      Wlroots Screencopy: ON
      Hyprland Toplevel Export: ON
  X11: ON
  I3/Sway: ON
    I3/Sway IPC: ON
  System Tray: ON
  PipeWire: ON
  Mpris: ON
  Pam: ON
  Polkit: ON
  Greetd: ON
  UPower: ON
  Notifications: ON
  Bluetooth: ON
  Network: ON

===== Runtime Information =====
Runtime Qt Version: 6.10.2
Crashed process ID: 159321
Run ID: ybh20cis9t
Shell ID: 271772659c0a416a1568f9aff43a625e
Config Path: /etc/xdg/quickshell/noctalia-shell/shell.qml

===== Report Integrity =====
Minidump save status: 0
Log save status: 0
Binary copy status: 0

===== System Information =====

/etc/os-release:
NAME="CachyOS Linux"
PRETTY_NAME="CachyOS"
ID=cachyos
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://cachyos.org/"
[...]

PR #567, opened 2026-02-13
services/pipewire: avoid use-after-free in PwDefaultTracker::reset()

bbedward:
relevant to #534

  `onFatalError()` is queued, so by the time `reset()` runs the PwNode objects may already be destroyed. The setter
  functions go through `setSimpleObjectHandle` → `QObject::disconnect()` on dangling pointers → SIGSEGV.

This changes to directly null the pointers and emit signals in `reset()`, skipping the disconnect-from-dead-objects path.

### DIFF
--- a/src/services/pipewire/defaults.cpp
+++ b/src/services/pipewire/defaults.cpp
@@ -12,7 +12,6 @@
 #include <spa/utils/json.h>
 
 #include "../../core/logcat.hpp"
-#include "../../core/util.hpp"
 #include "metadata.hpp"
 #include "node.hpp"
 #include "registry.hpp"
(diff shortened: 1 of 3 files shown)


outfoxxed:
Shouldn't this already be handled by the QObject::destroyed connection in SimpleObjectHandle? (I am not particularly proud of that interface, which should probably be replaced by bindables, but it should theoretically work for this)

bbedward:
You're right but somehow something causes mDefaultSink to point to freed memory in these crashes, but I dont see a case off hand why the destroyed signal wouldnt work - I'll look more into it.

bbedward:
Alternative fix pushed with per-handle connections, should still fix the crash - which I believe is caused by the nodes sharing the same destroyed slot.

outfoxxed:
Ah I know what this was.... I hate C++ class construction and teardown.

The class actually changes its type during destruction, as well as its this pointer, so they can't be compared. This used to be responsible for a bunch of QML incubator crashes before I rewrote that entirely. Your PR here should fix it.

src/services/pipewire/defaults.cpp
```diff
@@ -240,10 +213,18 @@ void PwDefaultTracker::setDefaultSink(PwNode* node) {
...
+
+	if (node) {
+		this->mDefaultSinkDestroyConn = QObject::connect(node, &QObject::destroyed, this, [this]() {
+			qCInfo(logDefaults) << "Default sink destroyed.";
+			this->mDefaultSink = nullptr;
+			emit this->defaultSinkChanged();
+		});
+	}
+
+	emit this->defaultSinkChanged();
```
outfoxxed, line 226: Prefer the strategy used in the rest of the project: create an onXDestroyed method, and disconnect by source->target with nullptr params after verifying the old object isn't null

outfoxxed (changes requested):
otherwise lgtm

outfoxxed:
thanks

PR #579, opened 2026-02-19
core/proxywindow: expose updatesEnabled property

bbedward:
Exposes `updatesEnabled` property from QQuickWindow 

Seems like QtQuick repaints unchanged surfaces constantly in its animation render loop, even if they are unchanged. This allows generally disabling repaints on surfaces.


outfoxxed (approved):
Yeah this is useful, the qtquick repainting choices are rather disappointing, especially because the engine knows exactly what regions changed.

PR #605, opened 2026-03-04
sway monitor fixes upstream

Mathew-D:
When monitors were turned off quick shell would crash. When it went into lock it would red screen. This fixes that.

This PR fixes two issues seen on Sway when monitors are powered off/on:

-     quickshell could segfault during output/workspace churn
-     after recovery, one monitor panel could remain missing

Root Cause

The i3/Sway IPC path was keeping cross-references between monitor/workspace objects that could become stale during rapid output/workspace updates.
Panel windows that are recreated on invisibility could lose requested-visible intent during compositor-driven hide/recreate cycles.

### DIFF
--- a/src/x11/i3/ipc/monitor.cpp
+++ b/src/x11/i3/ipc/monitor.cpp
@@ -74,7 +74,7 @@ void I3Monitor::setFocusedWorkspace(I3Workspace* workspace) {
 	}
 
 	this->bActiveWorkspace = workspace;
-};
+}
 
 void I3Monitor::onActiveWorkspaceDestroyed() { this->bActiveWorkspace = nullptr; }
 


src/x11/i3/ipc/controller.cpp
```diff
@@ -288,7 +288,9 @@ void I3IpcController::handleWorkspaceEvent(I3IpcEvent* event) {
 			qCInfo(logI3Ipc) << "Deleting" << oldWorkspace->bindableId().value() << name;
 
 			if (this->bFocusedWorkspace == oldWorkspace) {
-				this->bFocusedMonitor->setFocusedWorkspace(nullptr);
+				if (auto* focusedMonitor = this->bFocusedMonitor.value()) {
+					focusedMonitor->setFocusedWorkspace(nullptr);
+				}
 			}
```
outfoxxed, line 292: This is redundant with the destroy signal path.

src/x11/i3/ipc/monitor.cpp
```diff
@@ -55,6 +55,29 @@ void I3Monitor::updateFromObject(const QVariantMap& obj) {
 
 void I3Monitor::updateInitial(const QString& name) { this->bName = name; }
 
-void I3Monitor::setFocusedWorkspace(I3Workspace* workspace) { this->bActiveWorkspace = workspace; };
+void I3Monitor::setFocusedWorkspace(I3Workspace* workspace) { this->setActiveWorkspace(workspace); };
+
+void I3Monitor::setActiveWorkspace(I3Workspace* workspace) {
```
outfoxxed: Is there a reason to have one function that just calls the other one here?

outfoxxed (changes requested):
There are two separate changes here. In the future please make one PR per isolated change so unrelated changes aren't blocked on reviews of other changes.

src/x11/i3/ipc/controller.cpp
Mathew-D: Agreed this null-check looks redundant with the existing binding and destroy handling. I added it defensively while debugging and never removed it.

src/x11/i3/ipc/monitor.cpp
Mathew-D: I left it this way to stay consistent with activeWorkspace, and to keep this fix small without touching a bunch of call sites. Mostly just a naming thing.

src/x11/i3/ipc/controller.cpp
outfoxxed: No the whole if is redundant because the monitor will be updated by the destroy signal

src/x11/i3/ipc/monitor.cpp
outfoxxed: Just move the body to setFocusedWorkspace or delete setFocusedWorkspace

src/x11/i3/ipc/controller.cpp
```diff
@@ -288,9 +288,7 @@ void I3IpcController::handleWorkspaceEvent(I3IpcEvent* event) {
 			qCInfo(logI3Ipc) << "Deleting" << oldWorkspace->bindableId().value() << name;
 
 			if (this->bFocusedWorkspace == oldWorkspace) {
-				if (auto* focusedMonitor = this->bFocusedMonitor.value()) {
-					focusedMonitor->setFocusedWorkspace(nullptr);
-				}
+				this->bFocusedMonitor->setFocusedWorkspace(nullptr);
```
outfoxxed, line 291: The set is redundant as well

Mathew-D: Sorry about all the extra nullptr checks, I was mainly trying to catch the error that was crashing QS whenever I turned off my screens in Sway. I don’t currently run mainline QS anymore since the shell I use has moved to its own fork, and this patch has been working well there with no more crashes. Please feel free to adjust this PR as needed.

PR #643, opened 2026-03-18
i3/workspace: Use name instead of id for activate

badly-drawn-wizards:
Mirrors the change made for hyprland/workspace in f90bef2d994c88f075dbc2fcd81140e160351328 and does it for i3/sway.

### DIFF
--- a/changelog/next.md
+++ b/changelog/next.md
@@ -61,6 +61,7 @@ set shell id.
 - Fixed partial socket reads in greetd and hyprland on slow machines.
 - Worked around Qt bug causing crashes when plugging and unplugging monitors.
 - Fixed HyprlandFocusGrab crashing if windows were destroyed after being passed to it.
+- Fixed `I3Workspace.activate()` sending invalid commands to i3/sway for named or special workspaces.
 
 ## Packaging Changes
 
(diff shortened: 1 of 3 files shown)


outfoxxed:
Why? Does I3 support relative navigation? It's manual doesn't seem to indicate that it does.

badly-drawn-wizards:
> 6.8. Changing (named) workspaces/moving to workspaces
> 
> To change to a specific workspace, use the workspace command, followed by the number or name of the workspace. 
From https://i3wm.org/docs/userguide.html

`workspace ${ws.name}` does indeed function for both i3 and sway

For the same reason the hyprland change was made: named workspaces without a number prefix in their name have a number of `-1`, which fail to activate, whereas using the name always succeeds as the name uniquely identifies the workspace.

Not sure what you are referring to with relative navigation. This is about switching to workspace by name rather than number. If that is relative navigation, which I don't think is what you meant, then yes it is supported in i3/sway.

outfoxxed:
The reason for the hyprland change was that navigation to negative workspace ids triggered relative navigation instead. Just to clarify, *unnamed* i3 workspaces will still work with ${name}? Also are we supposed to string escape it?

badly-drawn-wizards:
I have tested with sway with named (with and without spaces) and unnamed workspaces. String escaping would be better, I was just following the hyprland change bar for bar.

I will further test with i3 to make sure that it behaves as sway does, but documentation suggests that it does.

PR #651, opened 2026-03-19
launch: add ability to override AppId via pragma or QS_APP_ID

bbedward:
Since quickshell can be used to build standalone desktop apps, in a way - I think it makes sense to be able to override the hardcoded app ID so if I have multiple quickshell-built applets I can have desktop entries for each, icons, and window/layer rules that are targeted to the specific app IDs.

The only question is whether it should also be overridden in the crash reporter, but I assume it should be so I just made it apply universally.

### DIFF
--- a/changelog/next.md
+++ b/changelog/next.md
@@ -38,6 +38,7 @@ set shell id.
 - Added `QS_DISABLE_FILE_WATCHER` environment variable to disable file watching.
 - Added `QS_DISABLE_CRASH_HANDLER` environment variable to disable crash handling.
 - Added `QS_CRASHREPORT_URL` environment variable to allow overriding the crash reporter link.
+- Added `AppId` pragma and `QS_APP_ID` environment variable to allow overriding the desktop application ID.
 
 ## Bug Fixes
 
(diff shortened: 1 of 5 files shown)


ashley-paul-12345:
> Since quickshell can be used to build standalone desktop apps, in a way - I think it makes sense to be able to override the hardcoded app ID so if I have multiple quickshell-built applets I can have desktop entries for each, icons, and window/layer rules that are targeted to the specific app IDs.
> 
> The only question is whether it should also be overridden in the crash reporter, but I assume it should be so I just made it apply universally.

can this also be done for FloatingWindows? so we can have a single instance of quickshell with FloatingWindows of different classes or intialClasses like for example in Hyprland?

outfoxxed:
> can this also be done for FloatingWindows

With some hacks and an intermediary period where its set to the global appid, yes. We'd need some changes in qt to be able to do it correctly.

> so if I have multiple quickshell-built applets I can have desktop entries for each, icons, and window/layer rules that are targeted to the specific app IDs.

I'm willing to accept the change generally but doing that is a bad idea for resource usage. Implementing a hack like I described above (see qwaylandwindow.cpp and search setAppId to see why its a problem) is probably a much better answer.

PR #672, opened 2026-03-28
core: improve named icon detection

charlesrocket:
`IconImageProvider` now picks up loose icons in `.local/share/icons`/etc (#532 rework).

### DIFF
--- a/src/launch/launch.cpp
+++ b/src/launch/launch.cpp
@@ -8,10 +8,10 @@
 #include <qhash.h>
 #include <qlist.h>
 #include <qlogging.h>
-#include <qnamespace.h>
 #include <qprocess.h>
 #include <qqmldebug.h>
 #include <qquickwindow.h>
+#include <qstandardpaths.h>
 #include <qstring.h>
 #include <qtenvironmentvariables.h>
 #include <qtextstream.h>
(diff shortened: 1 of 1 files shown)


src/core/iconimageprovider.cpp
```diff
@@ -7,14 +7,35 @@
...
+
 QPixmap
 IconImageProvider::requestPixmap(const QString& id, QSize* size, const QSize& requestedSize) {
 	QString iconName;
 	QString fallbackName;
 	QString path;
 
+	standardPaths();
+
 	auto splitIdx = id.indexOf("?path=");
```
outfoxxed, line 18: I think you actually just need to tack this onto launch.cpp#203

outfoxxed (changes requested):

src/launch/launch.cpp
```diff
@@ -196,6 +197,8 @@ int launch(const LaunchArgs& args, char** argv, QCoreApplication* coreApplicatio
 	{
 		QList<QString> dataPaths;
 
+		standardPaths();
+
```
outfoxxed: It looks like the whole function can just be reduced to adding a /icons line after /pixmaps

outfoxxed (changes requested):

src/launch/launch.cpp
charlesrocket: I would still have to add `.local/share`. Maybe I can just replace this with standard paths to avoid overloading with hardcoded values.

charlesrocket:
Just adding `/icons` misses some other paths. I replaced the old block with the standard path approach, so now all paths are covered:

```
DEBUG: Icon fallback path: "/home/charlie/.local/share/icons"
DEBUG: Icon fallback path: "/home/charlie/.local/share/pixmaps"
DEBUG: Icon fallback path: "/usr/local/share/icons"
DEBUG: Icon fallback path: "/usr/local/share/pixmaps"
DEBUG: Icon fallback path: "/usr/share/icons"
DEBUG: Icon fallback path: "/usr/share/pixmaps"
```

PR #682, opened 2026-03-29
Meson

ghost:
# Meson Build System

Imagine… a better world. A world without CMake.

# Status

- [x] Workarounds for missing Qt6 module features (qsb, static plugin resources, wayland protocol scanner, some private headers)
- [x] Checks and feature toggles for everything
- [x] More detailed post-config summary than usual
- [x] Fallback wraps for CLI11, cpptrace **and** tcmalloc while we’re at it
- [x] Test suite works (uses Test Anything Protocol)
- [x] PCH (they’re per-target in Meson though, so not worth it for small units)
- [ ] Installing instead of embedding resources. Should be 3 lines or so but I didn’t care yet.
- [ ] CI workflows
- [ ] End user docs
- [ ] Checks for things like Vulkan headers that should be pulled in by Qt anyway
- [ ] Tests with anything older than Qt 6.10.2 on anything other than Tumbleweed
- [ ] Sleep

# Design

Trying to do as much work as possible with the smallest amount of boilerplate and repetition. The ugly parts with the workarounds stay mostly in one place and should be easy to swap out once Meson catches up.

It is a bit hard to believe how terrible Qt’s CMake macros are. More than two thirds of all that code (many thousand lines) just deals with CMake’s shortcomings. I just don’t understand why all the corpos just keep forcing this absolute garbage on everyone.


NotAShelf (approved):
I have imagined a world without CMake for 3 seconds and my cortisol levels dropped significantly. LGTM

outfoxxed:
I briefly skimmed this and tried to build. The first thing I'm noticing is the qt component resolution does not support split qt packages, which means this is unusable for nix (and therefore our ci and a large number of users).

```
Detecting Qt6 tools
Run-time dependency qt6 (modules: Core) found: YES 6.10.2 (pkg-config)
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/bin/qmlcachegen found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/qmlcachegen found: NO
Program qmlcachegen6 found: NO
Program qmlcachegen-qt6 found: NO
Program qmlcachegen found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/bin/qmltyperegistrar found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/qmltyperegistrar found: NO
Program qmltyperegistrar6 found: NO
Program qmltyperegistrar-qt6 found: NO
Program qmltyperegistrar found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/bin/moc found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/moc found: YES 6.10.2 6.10.2 (/nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/moc)
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/bin/uic found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/uic found: YES 6.10.2 6.10.2 (/nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/uic)
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/bin/lrelease found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/lrelease found: NO
Program lrelease6 found: NO
Program lrelease-qt6 found: NO
Program lrelease found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/bin/rcc found: NO
Program /nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/rcc found: YES 6.10.2 6.10.2 (/nix/store/8ypf1x6g3hkk0c2sld72jfrmpk3aj8kv-qtbase-6.10.2/libexec/rcc)
Run-time dependency qt6 (modules: WaylandClient) found: YES 6.10.2 (pkg-config)
Has header "private/qwaylandinputdevice_p.h" with dependencies Qt6Core, Qt6WaylandClient: YES
Has header "QtWaylandGlobal/private/qtwaylandglobal-config_p.h" : YES
Run-time dependency wayland-client found: YES 1.24.0
Run-time dependency wayland-protocols found: YES 1.47
Run-time dependency xkbcommon found: YES 1.11.0
Program wayland-scanner found: YES (/nix/store/ym2nazjm9vqi0md5dw03vsh9zhg4v6q9-wayland-scanner-1.24.0-bin/bin/wayland-scanner)
Program qtwaylandscanner found: NO

src/wayland/meson.build:79:19: ERROR: Program 'qtwaylandscanner' not found or not executable
```

Also:
- this is 2.5k lines of buildscripts so at a bare minimum it would need to be in ci and kept in sync.
- I would rather not have buildscripts for dependencies here.
- Its very likely we'll have to generate cmake support code for users running in the planned library mode, unclear if that can usefully extend to meson.

ghost:
> split Qt packages

Yeah, Nix’ entire premise is to break the way things are expected to be packaged and then get mad when it doesn’t work. But jokes aside, that **is** a bug in Meson’s Qt module and it should be fixed regardless of what I think about it. I don’t think that’s going to be hard at all, and I want to do that, but I can’t guarantee that it will happen very soon because I use Tumbleweed and the solution I envision here, because of Nix’ unique approach and Meson’s philosophy to get the hell out of the way and just f… work, is to have first-class integration of Nix’ package management as its own dependency lookup method to fix the issue once and for all (including for non-Qt dependencies), rather than trying to hack the search paths until they just happen to work for a limited use case. Ideally, someone who knows and loves Nix would do it, but maybe this is one of those community jobs that just will not get done until there is a proof of concept.

>this is 2.5k lines of buildscripts so at a bare minimum it would need to be in ci and kept in sync.

That is the plan. Well, actually, the plan is also for me to actively contribute to Meson, make our build scripts shrink to a much smaller size, and eventually remove CMake once it’s mature enough.

>I would rather not have buildscripts for dependencies here.

Agreed. They should go into WrapDB so they can be maintained there and we only need to reference them. The alternative to that is relying on Meson’s built-in CMake support instead. This can be done with a minimal change to the .wrap files. I don’t like that solution because it mixes build systems and exposes us to CMake’s own set of problems. But at least it wouldn’t cause any additional ones: Meson runs CMake and extracts proper build targets from the generated output instead of relying on shell commands and prayers.

In any case, those are just fallbacks for when the dependencies cannot be found on the system, and won’t be used otherwise. I considered it worthwhile to have those scripts before submitting proper and complete versions to WrapDB because for our purposes the only thing to maintain is a list of source files.

>Its very likely we'll have to generate cmake support code for users running in the planned library mode, unclear if that can usefully extend to meson.

As a matter of fact, yes. Meson actually has pretty good integration of CMake and calls it directly to generate the necessary support files, so it would be trivial to support that and share any template files with the existing CMake build system.

My personal preference is to completely avoid doing that though, and exclusively support pkg-config, which does the same thing but better and with far less complexity. The CMake suffering in the free software community needs to end, and that can only happen if we stop actively supporting it and create better alternatives.

outfoxxed:
> I envision here, because of Nix’ unique approach and Meson’s philosophy to get the hell out of the way and just f… work, is to have first-class integration of Nix’ package management as its own dependency lookup method to fix the issue once and for all (including for non-Qt dependencies), rather than trying to hack the search paths until they just happen to work for a limited use case.

First off calling nix from inside nix is not going to work well, those kind of integrations have to be done at the nixpkgs level where its meson support is, but the good news is you can just use the same search path code qt's cmake modules expect, which is a : separated list in QT_ADDITIONAL_PACKAGES_PREFIX_PATH.

> That is the plan. Well, actually, the plan is also for me to actively contribute to Meson, make our build scripts shrink to a much smaller size, and eventually remove CMake once it’s mature enough.

I don't know how fast meson pulls in changes, but if possible I'd rather have the required ones merged upstream instead of hacked around temporarily here.

> The alternative to that is relying on Meson’s built-in CMake support instead.

I'd rather do this because it means I no longer have to worry about a dependency updating and breaking.

> My personal preference is to completely avoid doing that though, and exclusively support pkg-config, which does the same thing but better and with far less complexity. The CMake suffering in the free software community needs to end, and that can only happen if we stop actively supporting it and create better alternatives.

There are two parts needed for Quickshell to work as a library in the way I intend. One of them is easily solved with pkg-config, but the other is that it needs to output build definitions for a scanned configuration to be able to retain the ease of use it currently has while letting users who would be compiling anyway compile the config definitions in, which will give a few large configurations a noticeable speed boost.

Under cmake its fairly easy to just spit out a bunch of text for cmake to run, but I don't know how far this extends to meson.

ghost:
>First off calling nix from inside nix is not going to work well, those kind of integrations have to be done at the nixpkgs level where its meson support is, but the good news is you can just use the same search path code qt's cmake modules expect, which is a : separated list in QT_ADDITIONAL_PACKAGES_PREFIX_PATH.

Well that sounds simple enough.

>I don't know how fast meson pulls in changes, but if possible I'd rather have the required ones merged upstream instead of hacked around temporarily here.

Relatively quickly. It’s just a bit of a daunting task because proper support of course means considering a lot more edge cases, and the Qt CMake modules don’t cleanly separate CMake hacks (roughly 80% of the almost 100k lines of code) and actual logic so they’re a pain to study. Hence I was hoping to get something “good enough” here first, also to identify potential problems beforehand and have practical experience to guide the implementation.

>but the other is that it needs to output build definitions for a scanned configuration to be able to retain the ease of use it currently has while letting users who would be compiling anyway compile the config definitions in, which will give a few large configurations a noticeable speed boost.

If I’m understanding you correctly then that’s not hard to do at all.

PR #711, opened 2026-04-06
network: support ethernet devices

cpwrs:
Adds EthernetDevice(s) to Networking.devices with properties network, linkSpeed, and hasLink.

Adds NMWiredDevice and NMGenericNetwork to the NM backend. NMGenericNetwork mirrors a base frontend Network for all of the device types that only have one logical network. Backend objects now own the lifetime of their respective frontend object. This gets rid of the need to manage separate lists of frontend objects. NMNetwork is now the common base of NMWirelessNetwork and NMGenericNetwork.

### DIFF
--- a/src/network/network.hpp
+++ b/src/network/network.hpp
@@ -7,147 +7,14 @@
 #include <qtypes.h>
 
 #include "../core/doc.hpp"
-#include "../core/model.hpp"
-#include "device.hpp"
 #include "enums.hpp"
 #include "nm/settings.hpp"
 
 namespace qs::network {
+class NetworkDevice;
+}
 
-class NetworkBackend: public QObject {
-	Q_OBJECT;
-
-public:
-	[[nodiscard]] virtual bool isAvailable() const = 0;
-
-protected:
-	explicit NetworkBackend(QObject* parent = nullptr): QObject(parent) {};
-};
-
-class Networking: public QObject {
-	Q_OBJECT;
-
-public:
-	static Networking* instance();
-
-	void checkConnectivity();
-
-	[[nodiscard]] ObjectModel<NetworkDevice>* devices() { return &this->mDevices; }
-	[[nodiscard]] NetworkBackendType::Enum backend() const { return this->mBackendType; }
-	QBindable<bool> bindableWifiEnabled() { return &this->bWifiEnabled; }
-	[[nodiscard]] bool wifiEnabled() const { return this->bWifiEnabled; }
-	void setWifiEnabled(bool enabled);
-	QBindable<bool> bindableWifiHardwareEnabled() { return &this->bWifiHardwareEnabled; }
-	QBindable<bool> bindableCanCheckConnectivity() { return &this->bCanCheckConnectivity; }
-	QBindable<bool> bindableConnectivityCheckEnabled() { return &this->bConnectivityCheckEnabled; }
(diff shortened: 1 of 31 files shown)


outfoxxed (changes requested):
Everything seems to work, didn't spot anything wrong in the impl.

It would be nice if the ethernet/wifi network api had a bit more overlap. The main things I can think of are:
- expose DeviceType on network (makes delegate reuse easier between wifi/wired)
- expose networks[] on all devices, for wired its just always of length 1. Keep the network property though thats better if you know its wired
- keep name on network, just return the device name for ethernet

src/network/qml.hpp
cpwrs, line 1: Had to separate this - circular includes

src/network/nm/backend.cpp
```diff
@@ -146,9 +146,9 @@ void NetworkManager::registerDevice(const QString& path) {
 					QObject::connect(dev, &NMDevice::addAndActivateConnection, this, &NetworkManager::addAndActivateConnection);
 					QObject::connect(dev, &NMDevice::activateConnection, this, &NetworkManager::activateConnection);
 					// clang-format on
-
-					this->mDevices[path] = dev;
-					emit this->deviceAdded(dev->frontend());
+					QObject::connect(dev, &NMDevice::loaded, this, [this, dev]() {
+						emit this->deviceAdded(dev->frontend());
+					});
```
cpwrs, line 151: This was required to keep EthernetDevice.network a CONSTANT. Network.name depends on the device name dbus property loading (async)

outfoxxed: I'm not sure it needs to be a constant, and that also brings up an interesting question: Should an unplugged ethernet adapter have a network? It doesn't seem like it should to me. (networks[] also empty if we null this)

src/network/network.hpp
```diff
@@ -157,6 +20,10 @@ class Network: public QObject {
 	QML_UNCREATABLE("Network can only be aqcuired through networking devices");
 
 	// clang-format off
+	/// The name of the network.
+	Q_PROPERTY(QString name READ name CONSTANT);
+	/// The type of device this network belongs to.
+	Q_PROPERTY(DeviceType::Enum deviceType READ deviceType CONSTANT);
```
outfoxxed: Thinking about this, instead of giving the network the device type, we should probably give it the actual device pointer, which allows device.type

outfoxxed (changes requested):

src/network/nm/backend.cpp
cpwrs: Good point! I gated visibility behind hasLink. 

I kept deviceAdded connected to the loaded signal anyway. I think its better to add the frontend object after its properties have been updated from DBus.

outfoxxed:
Code looks fine from what I remember, Looks like eths can be renamed on the fly and that happens on my system when I plug devices in. Seeing if its an easy fix

edit: yes

src/network/nm/network.cpp
```diff
@@ -0,0 +1,308 @@
...
+						    emit this->requestActivateConnection(ref->path());
+					    }
+				    }
+				    delete call;
+			    }
+			);
+			return;
+		}
+		if (auto* apRef = this->referenceAp()) {
+			emit this->requestAddAndActivateConnection(NMSettingsMap(), apRef->path());
```
outfoxxed: New NMSettingsMap() here instead of the one we built earlier and old code doesn't do that. Regression?

PR #765, opened 2026-04-26
bluetooth: lift rfkill soft block when enabling a blocked adapter

Mic92:
bluez rejects Powered=true while PowerState is off-blocked, so the toggle was dead until the user ran `rfkill unblock bluetooth` by hand. Write an RFKILL_OP_CHANGE_ALL/RFKILL_TYPE_BLUETOOTH event to /dev/rfkill (writable via logind uaccess; same approach as bluez-qt, g-s-d, blueman) and defer the Powered write until bluez observes the unblock, since an immediate write still races bluez's own rfkill reader.


src/bluetooth/adapter.cpp
```diff
@@ -51,17 +58,58 @@ QString BluetoothAdapter::adapterId() const {
...
 }
 
+void BluetoothAdapter::onStateChanged() {
+	if (this->mPendingEnable && this->bState != BluetoothAdapterState::Blocked) {
+		this->setEnabled(true);
+	}
+}
+
+bool BluetoothAdapter::tryRfkillUnblock() {
+	auto fd = open("/dev/rfkill", O_WRONLY | O_CLOEXEC);
```
Mic92, line 91: qfile does the wrong open flags here, so we use open.

Mic92:
After `rfkill unblock`, bluez updates `PowerState` asynchronously, so the
   deferred `setEnabled(true)` can land in one of three states:

   | `[Policy] AutoEnable` | `PowerState` when retry runs | `Powered` | retry outcome |
   |---|---|---|---|
   | `true` (default ≥ 5.65) | `on` | `yes` | no-op |
   | `true`, mid-transition | `off-enabling` | `no` | writes `Powered=true`, redundant |
   | `false` | `off` | `no` | writes `Powered=true`, adapter turns on |

   The first row is what you'll see on most distros, which is why "unblock
   alone" appears sufficient there. The third row is why the explicit
   `Powered` write is still needed.

Mic92:
Wrong base — branch was cut from the noctalia-qs fork and dragged in its whole history. Reopening from a branch based on quickshell master.

Mic92:
New attempt: https://github.com/quickshell-mirror/quickshell/pull/766

PR #798, opened 2026-05-12
fix: Audio devices not detected after pipewire changes

### LINKED ISSUE #558: Audio devices not detected after pipewire changes
Related to AvengeMedia/DankMaterialShell#1332

After commit @8d19beb, DMS shell is no longer able to detect audio devices.

- Audio devices still appear correctly in the audio devices widget
- PipeWire itself is working normally
- No devices are detected by the shell

I have not yet checked which specific commit after 8d19beb introduced the issue, but the problem started occurring after one of the 3 services/pipewire changes.

Cloud0310:
Fixes #558.

This PR works around a type metadata issue caused by building Quickshell with `mold` and LTO enabled, which prevents Quickshell from correctly detecting PipeWire audio devices.

The issue occurs because `mold` strips required Qt QML type information during the LTO process, leading to type mismatches at runtime.

This workaround adds `[[gnu::visibility("default")]]` to preserve the required type metadata so it is not discarded by the linker during LTO. With the metadata retained, PipeWire audio devices are detected correctly again.

This is a temporary workaround until the underlying `mold` LTO issue is resolved upstream.

### DIFF
--- a/src/services/pipewire/peak.hpp
+++ b/src/services/pipewire/peak.hpp
@@ -8,15 +8,13 @@
 #include <qvector.h>
 
 #include "node.hpp"
+#include "qml.hpp"
 
 namespace qs::service::pipewire {
 
-class PwNodeIface;
 class PwPeakStream;
 
 } // namespace qs::service::pipewire
-
-Q_DECLARE_OPAQUE_POINTER(qs::service::pipewire::PwNodeIface*);
 
 namespace qs::service::pipewire {
 
@@ -24,7 +22,7 @@ namespace qs::service::pipewire {
 /// Tracks volume peaks for a node across all its channels.
 ///
 /// The peak monitor binds nodes similarly to @@PwObjectTracker when enabled.
-class PwNodePeakMonitor: public QObject {
+class QS_PIPEWIRE_EXPORT PwNodePeakMonitor: public QObject {
 	Q_OBJECT;
 	// clang-format off
 	/// The node to monitor. Must be an audio node.
--- a/src/services/pipewire/qml.hpp
+++ b/src/services/pipewire/qml.hpp
@@ -14,6 +14,12 @@
 #include "node.hpp"
 #include "registry.hpp"
 
+#if defined(__GNUC__) || defined(__clang__)
+#define QS_PIPEWIRE_EXPORT [[gnu::visibility("default")]]
+#else
+#define QS_PIPEWIRE_EXPORT
+#endif
+
 namespace qs::service::pipewire {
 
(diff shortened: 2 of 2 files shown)


kchibisov:
~Instead of this, can you try adding `PUBLIC` instead of `PRIVATE` to  https://github.com/quickshell-mirror/quickshell/blob/master/src/services/pipewire/CMakeLists.txt ?~

Nah, won't do it, thought it had some heuristic in mold, seems like just bug that should be reported upstream.

Maybe just disable(don't enable it, since I never enable it) LTO entirely if `mold` is detected, instead of fix like that?

Cloud0310:
> ~Instead of this, can you try adding `PUBLIC` instead of `PRIVATE` to https://github.com/quickshell-mirror/quickshell/blob/master/src/services/pipewire/CMakeLists.txt ?~
> 
> Nah, won't do it, thought it had some heuristic in mold, seems like just bug that should be reported upstream.
> 
> Maybe just disable(don't enable it, since I never enable it) LTO entirely if `mold` is detected, instead of fix like that?

I do not think this workaround should cause any issues. `visibility("default")` is already the normal/default visibility for most linkers, so this change should not negatively affect performance or behavior outside of this specific `mold` + LTO case.

Detecting `mold` and disabling LTO conditionally could also work, though I would still prefer keeping at least some LTO optimizations enabled if possible.

My plan is to keep this as a temporary workaround and add a comment in the code referencing the upstream `mold` bug report once I can produce a proper minimal reproduction.

I am still investigating the upstream issue. Linker and LTO interactions are fairly complex here, and I want to narrow down the exact condition that causes the Qt QML type metadata to be stripped before filing the report.

Cloud0310:
Your suggestion about detecting `mold` and disabling LTO conditionally is actually a pretty reasonable approach as well.

A few things I like about it:

* It keeps the workaround isolated to the exact problematic toolchain combination instead of touching symbol visibility in the code.
* It avoids relying on compiler/linker-specific attributes for runtime metadata preservation.
* It is easier to remove later once the upstream `mold` issue is fixed.
* It reduces the chance of accidentally affecting unrelated optimization behavior.

I still leaned toward this workaround because it preserves LTO for the affected target and fixes the runtime issue directly, though I am definitely open to switching approaches if we find a cleaner or more robust solution.

kchibisov:
The thing is, that other places could also be silently broken or break. In general, package should set lto, since forcing lto(or any fancy C flags) is generally not a great way and any sane packaging will just patch it out (unless it's proven that it works).

Issue #325, opened 2025-10-23
Audio volume level does not change when bluetooth device is connected


cyrneko:
this is closely related to this issue from dankMaterialShell, but also happens on noctalia and probably other shells: https://github.com/AvengeMedia/DankMaterialShell/issues/333

---

It seems that the second a bluetooth device is connected, audio adjustments other than `mute` cease to function

When changing the volume through `wpctl` or something like `pwvucontrol` the shell (and actual volume) changes accordingly.

because I've identified this with both DMS and noctalia I've decided to open an issue here, but I'd need some help for getting any relevant logs and data for diagnosing and fixing the bug

outfoxxed:
This might be fixed now in master

besseddrest:
just here to report that i'm still experiencing this issue

i just did an update today, i've got:

* dms-shell-bin 0.6.2-1
* quickshell-git 0.2.0.r44.ge9bad67-1

can follow up shortly with more details

PR #808, opened 2026-05-17
pipewire: write route volumes without volume step

### LINKED ISSUE #807: PwNodeAudio volume writes only update local state on Bluetooth sinks without volumeStep
## Summary

Setting `PwNodeAudio.volume` / `PwNodeAudio.volumes` on my default Bluetooth sink updates the QML-side value, but it does not change the PipeWire backend volume. Muting the same sink through `PwNodeAudio.muted` works, and changing the same sink through `wpctl set-volume` works.

This first showed up through DankMaterialShell master volume control, but I reduced it to Quickshell directly. DMS is doing the expected assignment to `Pipewire.defaultAudioSink.audio.volume`; the assignment just does not reach PipeWire for this sink.

## Environment

- Quickshell: `0.3.0` (Arch/CachyOS package)
- PipeWire: `1.6.5`
- WirePlumber: `0.5.14`
- Sink: Bluetooth / BlueZ output, `bluez_output.6C_5A_B5_EB_5E_8D.1` (`DingDong-124`)

There is also a local packaging warning that the package was built against Qt 6.11.0 while the system has Qt 6.11.1. I am including that for completeness, but the behavior below looks specific to the PipeWire volume write path rather than a crash or Qt issue.

## Minimal repro

With the Bluetooth sink as the default sink, `wpctl get-volume @DEFAULT_AUDIO_SINK@` initially reports:

```text
Volume: 0.40
```

Running this minimal Quickshell config changes the QML properties but not the actual backend volume:

```qml
import QtQuick
import Quickshell
import Quickshell.Services.Pipewire

ShellRoot {
    PwObjectTracker {
        objects: Pipewire.nodes.values.filter(node => node.audio && !node.isStream)
    }

    function logSink(prefix) {
        const sink = Pipewire.defaultAudioSink;
        console.log(prefix
            + " id=" + sink.id
            + " name=" + sink.name
            + " ready=" + sink.ready
            + " volume=" + sink.audio.volume
            + " muted=" + sink.audio.muted
            + " volumes=" + JSON.stringify(sink.audio.volumes));
    }

    Timer {
        interval: 1500
        running: true
        repeat: false
        onTriggered: {
            logSink("before");
            Pipewire.defaultAudioSink.audio.volume = 0.31;
            afterAverageTimer.start();
        }
    }

    Timer {
        id: afterAverageTimer
        interval: 1000
        repeat: false
        onTriggered: {
            logSink("after-average");
            Pipewire.defaultAudioSink.audio.volumes = [0.27, 0.27];
            afterChannelsTimer.start();
        }
    }

    Timer {
        id: afterChannelsTimer
        interval: 1000
        repeat: false
        onTriggered: {
            logSink("after-channels");
            Qt.quit();
        }
    }
}
```

Observed Quickshell log:

```text
before id=32 name=bluez_output.6C_5A_B5_EB_5E_8D.1 ready=true volume=0.4000000059604645 muted=false volumes=[0.4000000059604645,0.4000000059604645]
after-average id=32 name=bluez_output.6C_5A_B5_EB_5E_8D.1 ready=true volume=0.31 muted=false volumes=[0.31,0.31]
after-channels id=32 name=bluez_output.6C_5A_B5_EB_5E_8D.1 ready=true volume=0.27 muted=false volumes=[0.27,0.27]
```

But after the script exits, PipeWire still has the old volume:

```text
$ wpctl get-volume @DEFAULT_AUDIO_SINK@
Volume: 0.40
```

Directly changing the same sink works:

```text
$ wpctl set-volume @DEFAULT_AUDIO_SINK@ 0.35
$ wpctl get-volume @DEFAULT_AUDIO_SINK@
Volume: 0.35
```

Muting through Quickshell/DMS also works; only the volume write fails.

## Route detail

For this Bluetooth device route, `pw-cli enum-params <bluez-device-id> Route` contains `mute` and `channelVolumes`, but no `volumeStep`:

```text
mute false
channelVolumes Float 0.064000 0.064000
channelMap Id 3 4
```

On the same machine, internal/HDMI device routes do include `volumeStep`, for example:

```text
volumeStep Float 0.000015
```

So this may be specific to device-backed routes where PipeWire/WirePlumber does not expose a `volumeStep` field, which seems common for this BlueZ sink.

## Suspected cause

Looking at `PwNodeBoundAudio::setVolumes()` in `src/services/pipewire/node.cpp`, the device-backed branch appears to call `device->setVolumes(...)` only when `this->volumeStep != -1`. If `volumeStep == -1`, the backend write is skipped, but the function still assigns `mVolumes` and emits `volumesChanged()`. That matches the observed behavior exactly: QML state changes, but PipeWire remains unchanged.

`setMuted()` does not have the same `volumeStep` guard before calling `device->setMuted(...)`, which also matches the fact that mute works on the same sink.

Expected behavior would be for `volumeStep == -1` to still write the route volume, or at least not silently update the local QML state as though the backend accepted it.

faithleysath:
## Summary

Fix PipeWire route volume writes for devices that do not report `SPA_PROP_volumeStep`.

On these devices, `volumeStep` is parsed as `-1`. The previous logic only called `PwDevice::setVolumes(...)` from inside the `volumeStep != -1` branch, so route-backed sink volume writes updated Quickshell's local `mVolumes` state but did not write the new volume to PipeWire.

This keeps the existing minimum-step filtering when a device reports `volumeStep`, and treats a missing step value as "no step threshold", allowing the route volume write to proceed.

Fixes #807

## Testing

- `just configure debug -DNO_PCH=ON -DBUILD_TESTING=ON`
- `just build`
- `git diff --check`
- Tested locally against a Bluetooth sink (`bluez_output.6C_5A_B5_EB_5E_8D.1`) by writing `Pipewire.defaultAudioSink.audio.volumes = [0.24, 0.24]` from QML and verifying `wpctl get-volume @DEFAULT_AUDIO_SINK@` changed to `Volume: 0.24`.

`just lint-changed` was not run because this local checkout does not have the required `TIDYFOX` environment variable configured.

### DIFF
--- a/changelog/next.md
+++ b/changelog/next.md
@@ -1,4 +1,5 @@
 ## Bug Fixes
 
+- Fixed PipeWire route volume writes for devices without a reported volume step.
 - Fixed ScreencopyView not displaying when only lock surfaces are shown.
 - Fixed WlSessionLockSurface.visible crashing if accessed before backing surface creation.
--- a/src/services/pipewire/node.cpp
+++ b/src/services/pipewire/node.cpp
@@ -486,8 +486,8 @@ void PwNodeBoundAudio::setVolumes(const QVector<float>& volumes) {
 			                << "via device";
 			this->waitingVolumes = realVolumes;
 		} else {
+			auto significantChange = this->mServerVolumes.isEmpty() || this->volumeStep == -1;
 			if (this->volumeStep != -1) {
-				auto significantChange = this->mServerVolumes.isEmpty();
 				for (auto i = 0; i < this->mServerVolumes.length(); i++) {
 					auto serverVolume = this->mServerVolumes.value(i);
 					auto targetVolume = realVolumes.value(i);
@@ -496,25 +496,25 @@ void PwNodeBoundAudio::setVolumes(const QVector<float>& volumes) {
 						break;
 					}
 				}
+			}
 
-				if (significantChange) {
-					qCInfo(logNode) << "Changing volumes of" << this->node << "to" << realVolumes
-					                << "via device";
-					if (!this->node->device->setVolumes(this->node->routeDevice, realVolumes)) {
-						return;
-					}
-
-					this->mDeviceVolumes = realVolumes;
-					this->node->device->waitForDevice();
-				} else {
-					// Insignificant changes won't cause an info event on the device, leaving qs hung in the
-					// "waiting for acknowledgement" state forever.
-					qCInfo(logNode).nospace()
-					    << "Ignoring volume change for " << this->node << " to " << realVolumes << " from "
-					    << this->mServerVolumes
-					    << " as it is a device node and the change is too small (min step: "
-					    << this->volumeStep << ").";
+			if (significantChange) {
+				qCInfo(logNode) << "Changing volumes of" << this->node << "to" << realVolumes
+				                << "via device";
+				if (!this->node->device->setVolumes(this->node->routeDevice, realVolumes)) {
+					return;
 				}
+
+				this->mDeviceVolumes = realVolumes;
+				this->node->device->waitForDevice();
+			} else {
+				// Insignificant changes won't cause an info event on the device, leaving qs hung in the
+				// "waiting for acknowledgement" state forever.
+				qCInfo(logNode).nospace()
+				    << "Ignoring volume change for " << this->node << " to " << realVolumes << " from "
+				    << this->mServerVolumes
+				    << " as it is a device node and the change is too small (min step: " << this->volumeStep
+				    << ").";
 			}
 		}
 	} else {


ozwaldorf:
Hitting this issue on my machine as well, worked around using wpctl on bluetooth devices for now.

The following github issues are all reporting this bug:
- https://github.com/quickshell-mirror/quickshell/issues/325
- https://github.com/quickshell-mirror/quickshell/issues/807
- https://github.com/quickshell-mirror/quickshell/issues/857

ozwaldorf:
Confirmed this patch is working for me, applied to the latest 0.3.0 quickshell release via nix and tested on my configuration

ozwaldorf (approved):

nuknes:
hitting this on Arch with quickshell-git (0.3.1.r10, commit 2d3b3e9), Caelestia shell, HCH-RF02 (bluez5) as default sink

volume control is completely unresponsive on the BT sink; wpctl set-volume works fine on the same node

confirmed the root cause matches this PR: volumeStep parses as -1 for the bluez5 node, so PwDevice::setVolumes() is never called, only local mVolumes state updates

outfoxxed (approved):

PR #879, opened 2026-06-29
core/screen: retain removed screens to avoid use-after-free

sunshowers:
Previously, when a screen was disconnected, `QuickshellTracked::updateScreens` deleted its `QuickshellScreenInfo` wrapper. But a copy of that `QObject*` may have been put into a `QVariant` as a bare, unguarded pointer. Deleting the wrapper leaves that pointer dangling, and the next time the engine accesses it, the process segfaults with a use-after-free error.

This was observed as intermittent crashes on monitor wakeup under niri with DankMaterialShell, confirmed from two core dumps to be a freed `QuickshellScreenInfo` inside a `Toplevel.screens` list retained by a model.

Both core dumps had the following in them:

```
screens QVariant @ 0x7fecc79e4eb8  metatype='QList<QuickshellScreenInfo*>'
  inline QList {d=0x7fecf1921320, ptr=0x7fecf1921330, size=1}
    screens[0] = 0x7fed0313b6a0  <-- freed (vtable=0x1 d_ptr=0x7)
```

Fix this by not actually deleting lost screens and instead just putting them in an internal vector. Note that this is a memory leak, in the sense that the list will keep growing in an unbounded fashion. But over the lifetime of a quickshell process there aren't that many screens connected and disconnected, so it seems like a relatively straightforward way to handle this.

Note that `deleteLater` doesn't quite work here, because these raw pointers can be held across event loop iterations. A more structural fix is likely some kind of generational arena + weak index pattern, but that's outside the scope of this PR.

TODO:

- [ ] Is this the right fix? Any ideas on how to avoid or bound the memory leak?

### DIFF
--- a/src/core/qmlglobal.cpp
+++ b/src/core/qmlglobal.cpp
@@ -127,8 +127,10 @@ void QuickshellTracked::updateScreens() {
 	next:;
 	}
 
+	// Retain removed screens rather than deleting them, since users might hold on to
+	// a bare `QObject*` (via a `QVariant`) pointing at the old screen.
 	for (auto* oldScreen: this->screens) {
-		oldScreen->deleteLater();
+		this->mDeadScreens.push_back(oldScreen);
 	}
 
 	this->screens = newScreens;
--- a/src/core/qmlglobal.hpp
+++ b/src/core/qmlglobal.hpp
@@ -77,6 +77,11 @@ private slots:
 
 signals:
 	void screensChanged();
+
+private:
+	// Screens removed from the system are stored here instead of being
+	// deleted, since users might hold on to bare `QObject*`s pointing at them.
+	QVector<QuickshellScreenInfo*> mDeadScreens;
 };
 
 class QuickshellGlobal: public QObject {


sunshowers:
(I've patched this locally, and am going to run with it for a while to monitor if I see any other crashes.)

outfoxxed:
I've never managed to get the QML engine to choke on a freed pointer in any prior testing I've done to trigger this class of bug, but I went and did an extra round of testing and didn't find anything for similar cases.

There's really only one point we could reasonably return an invalid `QList<QuickshellScreenInfo*>` from in the backend which is `WindowsetProjection::screens`. I just fixed that in 1661ef10, but dms doesn't seem to use it so it may or may not help.

sunshowers:
Thanks. One thing to note here is that my monitor is a Samsung G9 57 which I run at 7680x2160 240Hz with an AMD GPU. I've seen that cause some issues on monitor wakeup, with link training failing for a bit then succeeding.

Both crashes I saw have this stack trace: https://gist.github.com/sunshowers/e439bd41d78fb5f189088471fc8d3e36

sunshowers:
> There's really only one point we could reasonably return an invalid QList<QuickshellScreenInfo*> from in the backend

From what I can tell it's not Quickshell returning a `QList<QuickshellScreenInfo*>` that is invalid at the time it is returned. Rather, Quickshell returns a _valid_ pointer that DMS caches internally. Then the pointer becomes invalid, and DMS attempts to access that.

sunshowers:
From my journalctl, both the crashes I saw have the following pattern:

```
21:48:47.980  kernel  amdgpu [drm] enabling link 0 failed: 15        <- DP link training fails
21:48:48.010  kernel  DM_MST: stopping TM on aconnector [id: 377]
21:48:48.047  niri    disconnecting connector: "DP-5"                <- output removed, screen wrapper freed
21:48:50.587  kernel  DM_MST: starting TM on aconnector [id: 377]
21:48:50.592  kernel  DP14, 4-lane link detected                     <- link training succeeds (~2.5s later)
21:48:51.218  kernel  Downstream port present 0, type 0
21:48:51.280  niri    new connector: DP-5 "Samsung Odyssey G95NC"    <- output re-added, screensChanged
21:48:51.280  niri    connecting connector: DP-5
21:48:51.281  niri    putting output DP-5 at x=0 y=0
21:48:51      ──────  crash (coredumpctl timestamp)                  <- re-wrap of the stale snapshot
```

So yeah, it seems like this is more likely to happen if a screen is added, removed, then rapidly added again afterwards. This might be why if you're not on a monitor that sometimes fails to do link training (such as a large DisplayPort MST display) you might not see the crash.

I'm not sure what facilities Quickshell or QML has to inject monitor add/remove events, but if there is such a mechanism, reproducing this might be more feasible.

sunshowers:
FWIW I've been running with my patch for a few days and it's been rock-solid -- previously I'd see around 2 crashes a day. Obviously not ideal to just leak memory but (from my limited understanding) I feel like the language/libraries/type system just doesn't provide enough knowledge for us to do better.

outfoxxed:
I had a chance to look at your stack trace. It looks like the QObject pointer is being laundered through a ScriptModel which explains the QVariant. I'll have to take a look at what DMS is doing but the gist of the problem is that updates to the model are non atomic and there are multiple potential ways to observe a dead pointer as is (including some dumb ones that just werent taken into consideration). Its not QScreen specific which is a good thing

outfoxxed:
One of the easy examples, more complex ones exist through the actual model interface including what DMS uses.
```qml
import QtQuick
import QtQuick.Controls
import QtQuick.Layouts
import Quickshell

FloatingWindow {
	ScriptModel {
		id: model
	}

	ColumnLayout {
		RowLayout {
			Button {
				text: "Update Model"
				onClicked: model.values = [...Quickshell.screens]
			}

			Button {
				text: "Update Usage Site"
				onClicked: v.text = `Model: ${model.values}`
			}
		}

		Text {
			text: `Screens: ${Quickshell.screens}`
		}

		Text {
			id: v
		}
	}
}
```

outfoxxed:
I have a patch for toplevel uafs and a strategy for embedded ones I think, Will give that a shot soon

sunshowers:
Thanks! Happy to run with a patch you propose for a while. Glad there's a better solution than the hacky one I have.

outfoxxed:
There's a good chance 4f2c62486b0c939f7fc1f7cea68037173813dbfe fixes it.

sunshowers:
Thanks! Running with that patch now.

PR #922, opened 2026-07-23
networking: add ability to manually add wifi networks

soramanew:
Currently, you can only connect to a wireless network that is exposed in the `NetworkDevice::networks` property. This PR adds the ability to connect to networks that do not expose their SSID, or that don't show up in the networks list for whatever reason.

This is done via a new `WifiDevice::addNetwork` method, which takes the SSID, security type and options, and whether or not the network is hidden.

Example usage:
```qml
Item {
    required property WifiDevice wifi
    
    Component.onCompleted: {
        wifi.addNetwork("some_ssid", WifiSecurityType.Wpa2Psk, {
            psk: "some_password"
        });
    }
}
```

### DIFF
--- a/src/network/nm/settings.cpp
+++ b/src/network/nm/settings.cpp
@@ -162,28 +162,9 @@ QVariantMap NMSettings::read() {
 }
 
 void NMSettings::write(const QVariantMap& settings) {
-	NMSettingsMap changedSettings;
 	NMSettingsMap removedSettings;
 	QStringList failedSettings;
-
-	for (auto it = settings.constBegin(); it != settings.constEnd(); ++it) {
-		if (!it.value().canConvert<QVariantMap>()) continue;
-
-		auto group = it.value().toMap();
-		QVariantMap toChange;
-		QVariantMap toRemove;
-		for (auto jt = group.constBegin(); jt != group.constEnd(); ++jt) {
-			if (jt.value().isNull()) {
-				toRemove.insert(jt.key(), QVariant());
-			} else {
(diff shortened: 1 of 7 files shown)


cpwrs:
This looks good and I have a few ideas/questions.

Is there a case for adding a network that isn't hidden?
We currently show a network as long as there exists an AP OR available connection for a given SSID.

I think we should add a `hidden` flag on `WifiNetwork`.

I'm working on ways to clean up the `NMSettings` mess so that we have a general settings class and don't have various maps to parse and validate. Hopefully that can end up replacing the credentials argument.

cpwrs:
> We currently show a network as long as there exists an AP OR available connection for a given SSID

Some work needs to be done to expose `org.freedesktop.NetworkManager.Settings` to list all networks including unavailable and hidden.

Then adding an `available` boolean on `Network` and/or a list on the `NetworkDevice` will make it clear which networks have found APs.

soramanew:
> Is there a case for adding a network that isn't hidden?

Not really, more so just to provide the option for it. E.g. in my shell there is an option to add a network manually, with a hidden toggle. So it needs to be able to specify hidden/not hidden. Android also has this.

> We currently show a network as long as there exists an AP OR available connection for a given SSID.
> 
> I think we should add a `hidden` flag on `WifiNetwork`.

This is not a bad solution, we can do that too.

> I'm working on ways to clean up the `NMSettings` mess so that we have a general settings class and don't have various maps to parse and validate. Hopefully that can end up replacing the credentials argument.

That sounds great. How would that be shaped though? There are a lot of different possible shapes depending on the credential type, etc

PR #930, opened 2026-07-28
fix: stop watching parent directories in DesktopEntryMonitor

### LINKED ISSUE #757: DesktopEntries fires spurious rescan events, growing log.qslog by multiple GB
> _This issue was drafted by Claude (an AI coding assistant) on behalf
> of the user._

## Summary

On Fedora 43 with stock `quickshell` (DMS config), `DesktopEntries` fires
spurious "Directory change detected, performing full rescan" events at
roughly **3 Hz** while no application directory is actually changing.
The resulting log growth fills the `/run/user/$UID` tmpfs:

- `qs` consumes ~**45% of one CPU** continuously at idle (DMS + stock
  quickshell, no user activity).
- `log.qslog` grew to **3.3 GB in ~16 h** under
  `/run/user/$UID/quickshell/by-id/<id>/`, enough to fill the 3.2 GB
  `tmpfs` and DoS everything else that writes under `$XDG_RUNTIME_DIR`.
- The log is almost entirely this repeating block (roughly every
  ~0.3 s):
  ```
  quickshell.desktopentry
  Directory change detected, performing full rescan
  Found desktop entry "aide"
  Found desktop entry "Alacritty"
  …<every .desktop file on the system>…
  ```
- First 20 MB of one session contained **771** "Directory change
  detected" lines.
- A fresh measurement after rebuilding (see below): **21 rescans in 10 s
  with zero real filesystem events** on any watched path.

## The events are spurious

Reporter watched all standard desktop-entry directories with
`inotifywait -mr -e modify,create,delete,moved_to,moved_from,attrib` for
8 seconds: **zero events** on `/usr/share/applications`,
`/usr/local/share/applications`, `~/.local/share/applications`,
`/var/lib/flatpak/exports/share/applications`, and
`~/.local/share/flatpak/exports/share/applications`. Over the same
interval, `qs` logged "Directory change detected" ~21 times and grew its
log by ~500 KB. The desktop-entry set has not changed in any
externally-observable way.

`qs` holds many `anon_inode:inotify` fds; the watch descriptors show
`mask:cc6` and `mask:7c4` (the usual modify/create/delete/attrib
bundles), so the issue isn't a mis-set event mask — some other signal is
causing the rescan code path to fire.

## Environment

- OS: Fedora 43
- quickshell: `0.2.1.1+pin713.26531fc-1.fc43` (upstream commit
  `26531fc46ef17e9365b03770edd3fb9206fcb460`)
- Originally installed from Copr `avengemedia/danklinux`; then rebuilt
  locally (see next section).
- Master HEAD at time of filing: `783c953987dc56ff0601abe6845ed96f1d00495a`.
  Searched the `26531fc46..783c9539` range for relevant fixes — nothing
  found that touches `DesktopEntries`'s rescan logic. Happy to re-test
  against master if useful.
- Qt: `qt6-qtbase-6.10.3-1.fc43`
- Compositor/shell: Niri + DankMaterialShell. DMS does **not** implement
  its own file watcher — it only reads
  `DesktopEntries.applications.values` (verified at
  `/usr/share/quickshell/dms/Services/AppSearchService.qml`). The loop
  lives in quickshell.

## Ruled out: Qt ABI mismatch

The Copr binary emitted quickshell's own "COMPATIBILITY WARNING:
Quickshell was built against Qt 6.10.1 but the system has updated to Qt
6.10.3…" banner. Reporter rebuilt the package locally against Fedora 43's
current Qt 6.10.3:

- `rpmbuild --rebuild` of the same SRPM (commit `26531fc46`).
- New binary's session log no longer prints the compatibility banner.
- CPU, log growth rate, rescan frequency are **unchanged** vs. the
  ABI-mismatched build. Bug is independent of ABI drift.

## Not a dup

- `quickshell-mirror/quickshell#468` (closed not-planned) was triaged as
  a Caelestia-specific `FileSystemModel` use-after-free. Log pattern is
  similar but the loop here is in stock quickshell `DesktopEntries`, not
  in a shell's QML. Stack trace there showed
  `caelestia::models::FileSystemModel::compareEntries()`; reporter here
  uses DMS, which has no such wrapper.
- `caelestia-dots/shell#1053` is the companion report for that same
  Caelestia bug.

## Ask

1. Any known conditions under which `DesktopEntries`'s
   `QFileSystemWatcher` reports spurious directory changes? (e.g. kernel
   ≥ 6.x inotify quirks, btrfs/ext4 dir-mtime differences, glibc
   `readdir` ordering, …)
2. Would a guard against "rescan found zero added/removed/changed
   entries" be acceptable upstream? That would bound the damage without
   changing watcher behaviour. Precedent in-tree: commit `9a9c605`
   ("core: hash scanned files and don't trigger a reload if matching",
   changelog entry *"Reloads are prevented if no file content has
   changed"*) applies exactly this pattern to QML config reload via
   `QmlScanner::hasFileContentChanged()`. The same shape on
   `DesktopEntries` would fix this.
3. Would you accept an env escape hatch (e.g.
   `QS_DISABLE_DESKTOP_ENTRY_WATCHER`)? Note that the existing
   `QS_DISABLE_FILE_WATCHER` (added in `e32b909`) only gates
   `QuickshellSettings::watchFiles()` — i.e. QML config reload — not
   `DesktopEntries`, so it's not applicable here. Reporter is willing
   to PR either option.

Happy to run any diagnostics, capture `-vv` output, or enable a
quickshell-side debug build.

short-circuit:
### Problem

`DesktopEntryMonitor::startMonitoring()` calls `addPathAndParents()`, which adds `QFileSystemWatcher` watches on every parent directory of each desktop entry path — all the way up to `/`. For example, for `/usr/share/applications` it watches `/usr/share`, `/usr`, and `/`.

Watching `/` means `QFileSystemWatcher::directoryChanged` fires on **any filesystem event anywhere on the system** — a temp file in `/tmp`, a cache write, a `.config` change — triggering a full rescan of all desktop entries. This causes:

- Spurious rescans at ~3 Hz even when no `.desktop` file has changed
- 45%+ CPU at idle
- `log.qslog` growing to multiple GB from "Directory change detected" spam
- Fill tmpfs under `$XDG_RUNTIME_DIR` (reported to fill 3.2 GB in ~16 h)

### Fix

Replace `addPathAndParents()` with a plain `watcher.addPath(path)` so only the actual desktop-entry directories are watched. `scanAndWatch()` independently handles subdirectory monitoring.

### Result

Measured on NVIDIA RTX 4090, Hyprland (Wayland), Arch Linux, Qt 6.11, ~366 .desktop files:

| Before | After |
|--------|-------|
| 45–65% CPU at idle | ~19% CPU (scene graph sync, no rescans) |
| 1–3 rescans per second | 0 rescans (initial scan only) |
| Log grows to GB | Clean log |
| Every filesystem event triggers rescan | Only `.desktop` dir changes trigger rescan |

### Related

Closes #757 — "DesktopEntries fires spurious rescan events, growing log.qslog by multiple GB"

### DIFF
--- a/src/core/desktopentrymonitor.cpp
+++ b/src/core/desktopentrymonitor.cpp
@@ -46,7 +46,7 @@ DesktopEntryMonitor::DesktopEntryMonitor(QObject* parent): QObject(parent) {
 void DesktopEntryMonitor::startMonitoring() {
 	for (const auto& path: DesktopEntryManager::desktopPaths()) {
 		if (!QDir(path).exists()) continue;
-		addPathAndParents(this->watcher, path);
+		this->watcher.addPath(path);
 		this->scanAndWatch(path);
 	}
 }


outfoxxed:
This would break any system that uses symlinks and that is not what the original code does.

If you can provide an inotify trace of events occurring on / we may be able to do something.

mortang2410:
Today I woke up to 6 GB of tmpfs being filled.

After a lot of digging, the culprit is quickshell because it literally adds parent directories to watch and constantly rescans and logs even if only metadata has changed for a direct subdir of ~ (which should not be relevant for desktop entry rescan).

I do not believe this behavior is reasonable or sane. There should be a better way to accommodate symlinks  (config settings, manual intervention etc.) rather than this. 

image

PR #949, opened 2026-08-11
niri/ipc: add niri IPC module

vk-hr:
Adds a Quickshell.Niri module with native C++ integration with niri's IPC socket, following the existing Hyprland module pattern (ObjectModel, bindable properties, singleton).

Covers workspaces, windows, monitors, screencasts, overview state, keyboard layouts, action dispatch, and a raw event stream. All niri IPC events are handled. 

Known limitation: there is no wayland Toplevel to niri window association, since niri doesn't expose window ids over wayland. Deferred pending an upstream niri mechanism.

To test: qs -p src/wayland/niri/test/manual/workspaces.qml (also windows.qml, monitors.qml, misc.qml)


blackd:
I would really love this accepted.

PR #958, opened 2026-08-16
networking: add traffic statistics to NetworkDevice

cpwrs:
Adds these properties to `NetworkDevice`:
- `rxBytes`: Number of received bytes
- `txBytes`: Number of transmitted bytes
- `refreshRate`: How often to refresh `rxBytes` and `txBytes` in milliseconds

Useful for upload/download speed

### DIFF
--- a/src/network/device.cpp
+++ b/src/network/device.cpp
@@ -6,6 +6,7 @@
 #include <qobject.h>
 #include <qstring.h>
 #include <qtmetamacros.h>
+#include <qtypes.h>
 
 #include "../core/logcat.hpp"
 #include "enums.hpp"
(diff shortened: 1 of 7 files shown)


outfoxxed:
What do the traffic statistics actually mean in absence of a refresh period? Seems like deriving any useful metric about speed would be hard unless theres an expected way to do that. When we do have a time frame, do you think normalizing it to bytes/sec seems like a good idea?

cpwrs:
> Seems like deriving any useful metric about speed would be hard unless theres an expected way to do that. 

Yeah.. this is a bad API from NM. I didn't consider that you can't count the # of refreshes and calculate speed from that. There's no update signal to count if no bytes are rx/txed.

> When we do have a time frame, do you think normalizing it to bytes/sec seems like a good idea?

Yes that'd be better. I could add an internal timer however you can't easily ensure it's synced with NMs. Polling /sys is easy, more accurate, and needed for other backends to support speed anyway.

outfoxxed:
Rather than keeping a separate timer you could just use rate to multiply bytes if nonzero, and probably just 0 out tx/rx if rate is 0

cpwrs:
Oh I should've clarified rx/tx bytes are a counter of total bytes since device init. 

I have to compute the delta which makes a period with no update (+0 bytes) an issue.
A timer would handle this scenario. This is how plasma implements it

outfoxxed:
A timer sounds fine but if we're doing any kind of polling measurement here we probably want it via a monitor component. I guess exposing the totals to qml lets that be done in qml.

PR #988, opened 2026-08-24
networking: add frequency, bssid, and other details to WifiNetwork

detluck:
### Summary
This PR exposes additional 802.11 access point properties from NetworkManager to `Quickshell.Networking.WifiNetwork`. These properties allow desktop shells to filter networks by frequency band (2.4 GHz vs 5 GHz), and display router MAC addresses and maximum link speeds in network detail views without spawning external CLI tools.

### Added Properties to `WifiNetwork`
* **`frequency`** (`quint32`): The radio frequency in MHz (useful for determining 2.4 GHz, 5 GHz, or 6 GHz bands).
* **`bssid`** (`QString`): The hardware MAC address of the access point.
* **`maxBitrate`** (`quint32`): Maximum bitrate supported by the access point in Kbps.
* **`bandwidth`** (`quint32`): Channel bandwidth announced by the access point in MHz (e.g. 20, 40, 80, 160).
* **`lastSeen`** (`qint32`): Timestamp in `CLOCK_BOOTTIME` seconds when the access point was last seen in scan results.

### DIFF
--- a/src/network/nm/network.cpp
+++ b/src/network/nm/network.cpp
@@ -204,6 +204,12 @@ NMWirelessNetwork::NMWirelessNetwork(const QString& ssid, NetworkDevice* device,
 	QObject::connect(this, &NMWirelessNetwork::referenceApChanged, this, updateSecurity);
 	QObject::connect(this, &NMWirelessNetwork::settingsRemoved, this, checkDisappeared);
 	QObject::connect(this, &NMWirelessNetwork::apRemoved, this, checkDisappeared);
+	QObject::connect(
+	    this,
+	    &NMWirelessNetwork::activeApPathChanged,
+	    this,
+	    &NMWirelessNetwork::updateReferenceAp
+	);
 
 	// Register and bind the frontend WifiNetwork.
 	this->mFrontend = new WifiNetwork(ssid, device, this);
@@ -215,6 +221,11 @@ void NMWirelessNetwork::updateReferenceAp() {
 	if (this->mAccessPoints.isEmpty()) {
 		this->bReferenceAp = nullptr;
 		this->bSignalStrength = 0;
+		this->bFrequency = 0;
+		this->bMaxBitrate = 0;
+		this->bHwAddress = QString();
+		this->bBandwidth = 0;
+		this->bLastSeen = QDateTime();
 		return;
 	}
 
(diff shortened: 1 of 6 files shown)


src/network/nm/network.cpp
```diff
@@ -260,6 +260,26 @@ void NMWirelessNetwork::bindFrontend() {
...
+		return ap ? ap->maxBitrate() : 0;
+	});
+	frontend->bindableBandwidth().setBinding([this]() {
+		auto* ap = this->referenceAp();
+		return ap ? ap->bandwidth() : 0;
+	});
+	frontend->bindableLastSeen().setBinding([this]() {
+		auto* ap = this->referenceAp();
+		return ap ? ap->lastSeen() : -1;
+	});
```
cpwrs: I don't think these nullptr checks should happen in `bindFrontend`.

I'd follow the existing pattern that `bSignalStrength` uses. 
Expose a bindable for each of these properties and bind/reset them in 
`updateReferenceAp()`.

src/network/nm/network.cpp
cpwrs, line 202: `updateReferenceAp` is called when an AP is added/removed or one of their signal strengths changes. We should also call it when the active path changes. 

`QObject::connect(this, &NMWirelessNetwork::activeApPathChanged, this, &NMWirelessNetwork::updateReferenceAp);`

cpwrs (changes requested):
Left a couple comments. Could you add the properties to the manual test as well?
Looks good besides that. 

We planned on eventually adding full AP exposure. For example with an `AccessPoint` class, `WifiNetwork.accessPoints` list and `WifiNetwork.activeAccessPoint` ptr.

I'm OK with getting these properties merged though. They could still exist as convenient accessors after the full AP API is added.

src/network/wifi.hpp
```diff
@@ -23,6 +23,16 @@ class WifiNetwork: public Network {
...
+	/// The radio frequency of the access point in MHz.
+	Q_PROPERTY(quint32 frequency READ default NOTIFY frequencyChanged BINDABLE bindableFrequency);
+	/// The hardware address (BSSID) of the access point.
+	Q_PROPERTY(QString bssid READ default NOTIFY bssidChanged BINDABLE bindableBssid);
+	/// The maximum bitrate this access point is capable of, in kilobits/second.
+	Q_PROPERTY(quint32 maxBitrate READ default NOTIFY maxBitrateChanged BINDABLE bindableMaxBitrate);
+	/// The channel bandwidth announced by the access point in MHz (e.g. 20, 40, 80, 160).
+	Q_PROPERTY(quint32 bandwidth READ default NOTIFY bandwidthChanged BINDABLE bindableBandwidth);
+	/// The timestamp (in CLOCK_BOOTTIME seconds) for the last time the access point was found in scan results.
+	Q_PROPERTY(qint32 lastSeen READ default NOTIFY lastSeenChanged BINDABLE bindableLastSeen);
```
cpwrs: We have the util `clockBootTimeToDateTime`. Might be nice to make this property a QDateTime. I think those become a JS Date after crossing the QML border, not sure.

cpwrs (commented):
Works for me! I'm getting 0 (unknown) for every bandwidth, but that's probably my driver or backend.

src/network/wifi.hpp
detluck: Right, great idea! I’ll do that soon. According to the Qt documentation:

> “The QML engine provides automatic type conversion between QDateTime values and JavaScript "Date" objects.”»

cpwrs (approved):

PR #990, opened 2026-08-24
build: add systemd services

LaserEyess:
This is a modification to #159 addressing the primary concern of being able to use multiple configuration files.

1. For "normal" use with `${XDG_CONFIG_HOME}/quickshell/shell.qml` the `quickshell.service` file is enough
2. To use an alternative config in `${XDG_CONFIG_HOME}/quickshell` you can use `quickshell@.service` which will pass the indentifier to `quickshell -c`
3. If that is still not enough you can use `quickshell-path@.service` which will (when properly escaped!) set `quickshell -p` to the absolute config of your choice.

The motivation behind this is simple: systemd lets you place quickshell in the session slice, which gives it special privileges against other processes running in your user session. In particular, it will shield quickshell against being considered for OOM killing (with systemd-oomd) as well as give it preferential CPU time when the system is slowed down by something else.

If any of this should be documented anywhere I'd be happy to do that as part of this PR, please let me know where you'd like that. I also considered deduplicating the services themselves and using `configure_file` more judiciously.


UnnaturalTwilight:
I'm not a dev but a quick thing I noticed is that the unit description lines should include "Quickshell" because systemd uses the description as a replacement for the name for things like journal logs.

From the systemd.unit.5 man page
> Description=\
A brief, meaningful, human-readable text identifying the unit. This may be used by systemd (and suitable UIs) as a user-visible label for the unit, so this string should identify the unit rather than just describe it, despite the name
 https://man.archlinux.org/man/systemd.unit.5#%5BUNIT%5D_SECTION_OPTIONS

PR #993, opened 2026-08-24
networking: fix omitted mode setting case in NM

### LINKED ISSUE #987: NetworkManager Wi-Fi profiles with an omitted mode are not marked known or connected
### Summary

An active NetworkManager Wi-Fi connection is not marked as `known` or `connected` by `Quickshell.Networking` when its saved profile omits `802-11-wireless.mode`.

NetworkManager treats an omitted mode as `infrastructure`, so this is a valid station profile. I reproduced this with an eduroam WPA2-Enterprise profile generated by GÉANT CAT.

### Environment

- Quickshell: `0.3.0`
- Revision: `28771c7c74b42e20afca0b1b63980cb46515537c`
- Package: `quickshell-git`
- NetworkManager: `1.58.0-1`
- Linux: `7.1.8-arch1-3`

### Steps to reproduce

1. Connect using a NetworkManager Wi-Fi profile whose mode is omitted.
2. Confirm that the profile is active and the mode is unset:

   ```console
   $ nmcli -g GENERAL.STATE,GENERAL.CONNECTION device show wlp3s0
   100 (connected)
   eduroam

   $ nmcli -f 802-11-wireless.mode connection show eduroam
   802-11-wireless.mode: --
   ```

3. Enable scanning and inspect the device and network through `Quickshell.Networking`.

### Actual result

The device is connected, but the active network is neither connected nor known:

```text
DEVICE wlp3s0 type 1 state 2 connected true
NETWORK "eduroam" state 4 connected false known false signal 0.6
```

Setting the profile's mode explicitly changes the Quickshell result:

```console
nmcli connection modify eduroam 802-11-wireless.mode infrastructure
```

```text
NETWORK "eduroam" state 2 connected true known true signal 0.6
```

NetworkManager remained connected during this test. Restoring the mode to an empty value reproduced the original result:

```console
nmcli connection modify eduroam 802-11-wireless.mode ''
```

### Expected result

An omitted or empty `802-11-wireless.mode` should be treated as `infrastructure`, matching NetworkManager's documented behavior:

> Wi-Fi network mode; one of "infrastructure", "mesh", "adhoc" or "ap". If blank, infrastructure is assumed.

<https://networkmanager.dev/docs/api/latest/settings-802-11-wireless.html>

### Probable cause

`NMWirelessDevice::onSettingsLoaded()` associates settings with a network only when the mode is explicitly `"infrastructure"`:

```cpp
const auto mode = map["802-11-wireless"]["mode"].toString();

if (mode == "infrastructure") {
```

<https://github.com/quickshell-mirror/quickshell/blob/28771c7c74b42e20afca0b1b63980cb46515537c/src/network/nm/wireless.cpp#L124-L138>

Profiles relying on NetworkManager's default empty value are therefore excluded before their settings and active connection can be associated with the scanned network.

### Possible fix

Treat an empty mode as infrastructure while preserving the current exclusion of explicit `ap`, `adhoc`, and `mesh` profiles:

```diff
-if (mode == "infrastructure") {
+if (mode.isEmpty() || mode == "infrastructure") {
```

cpwrs:
Fixes #987. In NetworkManager if the "mode" setting is omitted the default is infrastructure.

Thanks @Awlaursen

### DIFF
--- a/src/network/nm/wireless.cpp
+++ b/src/network/nm/wireless.cpp
@@ -124,7 +124,8 @@ void NMWirelessDevice::onSettingsLoaded(NMSettings* settings) {
 	const auto ssid = map["802-11-wireless"]["ssid"].toString();
 	const auto mode = map["802-11-wireless"]["mode"].toString();
 
-	if (mode == "infrastructure") {
+	// Omitted mode is assumed to be "infrastructure".
+	if (mode.isEmpty() || mode == "infrastructure") {
 		auto* net = this->mNetworks.value(ssid);
 		if (!net) net = this->registerNetwork(ssid);
 		net->addSettings(settings);


cpwrs:
This made me realize people could change the ssid or mode at any time. Those are the only ones that would warrant placing the settings under a different network but it should be handled. I'll fix in a separate PR.

Issue #445, opened 2025-12-28
Question: Plans for translation (i18n) support in Quickshell?


AhmedSaadi0:
Hi,

I noticed that while `qsTr()` calls exist in QML files, there currently seems to be no way to load translation files (`.qm`) for the application.  

I wanted to ask if there are any plans to support translations (i18n) in Quickshell?  
Is this something expected in the near future, or is there a roadmap for implementing it?  

Thank you.

sadorowo:
I think that this feature is needed for a lot of developers, I want to migrate from AGS but for now I'd need to write my own JSON-based translation system like I used to do in Aylur's GTK Shell. I want to take advantage of the powerful Qt multilanguage system and Qt Linguist app.

AhmedSaadi0:
Nobody has answered about the roadmap yet. I think I am going to make a Pull Request to add this feature myself. But honestly, I am too busy right now. I can't work on it soon. I will check the project later when I have time.

komagata:
Quickshell is a framework for building complete desktop shells, and localization seems like a capability that is best provided by the runtime rather than reimplemented by every shell configuration.

QML already provides the standard `qsTr()` family of functions, and the Qt translation ecosystem provides TS/QM catalogs, Qt Linguist, contexts, disambiguation, plural handling, and dynamic retranslation. However, a Quickshell configuration currently has no supported way to install and retain a `QTranslator`, so it cannot use that standard workflow end to end.

As a result, every localized Quickshell project must either build a native extension or create its own translation runtime and API. That fragments translation tooling and requires each project to solve the same locale, plural, context, and retranslation problems independently.

First-class catalog loading in Quickshell would let configurations use the Qt APIs already present in QML. Quickshell would not need to own any translations or decide how individual projects organize them; it would only provide the missing runtime integration for loading, replacing, and retaining translators safely across reloads.

Omarchy is one concrete example. During an AI-assisted investigation of its i18n support, I verified that an external QML module can install a `.qm` catalog and call `QQmlEngine::retranslate()` to update existing `qsTr()` bindings inside an unmodified Quickshell process. This confirms that the required Qt primitives work inside an unmodified Quickshell process, although the correct upstream API, translator lifetime, and reload behavior still need to be designed within Quickshell.

I am not publishing or submitting a Quickshell patch from that investigation. I’m only reporting the observed behavior to help clarify the required integration; any upstream implementation should be independently designed and written in accordance with Quickshell’s contribution policy.

[...]

PR #1081, opened 2026-09-05
core: load shell translation catalogs

komagata:
## Summary

Allow applications built with Quickshell to translate their UI using Qt's standard translation system. For example, an Omarchy plugin's strings can be included in the shell's translation catalog.

Without built-in catalog loading, applications need extra integration, such as a C++ QML extension that installs a `QTranslator`. I tried this in an earlier PoC, but it leaves the application responsible for building and distributing the extension and handling engine reloads. This change handles catalog loading in Quickshell itself.

Load `i18n/qml_<language>.qm` beside the root QML file, following the `QQmlApplicationEngine` naming convention. Applications can use `qsTr()` and other standard Qt translation APIs, with Qt handling plural forms and translation contexts. Set `Qt.uiLanguage` to switch languages.

One `QTranslator` is owned by `RootWrapper` across engine reloads, so catalogs from overlapping generations do not mix. Language changes replace the catalog and call `retranslate()`. Reloads preserve the selected language and load translations before QML object creation. Failed QML compilation leaves the running shell's translator intact.

Related to #445.

## Testing

- Built with Qt 6.11.2 and Clang 22.1.8, with tests enabled, PCH disabled and the crash handler disabled.
- Added one integration test covering startup translation, Russian plural forms, language switching and source-text fallback.
- The full suite passes 9 of 10 test targets. `popupwindow::moveWithParent` fails with the offscreen platform; the same failure was reproduced with the original source during earlier verification.
- Formatting checks and clang-tidy on the new test pass. Earlier comparison of clang-tidy diagnostics for `rootwrapper.cpp` found identical warnings in unchanged headers before and after this change.
- Qt 6.6, Nix, upstream CI and tidyfox have not been verified locally.

Test QML and translation data are separate fixtures. Qt Linguist Tools (`lrelease`) are required for test builds only; no runtime dependency is added.

## Scope

One catalog per shell. Updating a catalog requires an explicit shell reload. Per-plugin catalogs, RTL layout and OS locale settings are outside this change.

Developed with AI assistance.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -58,6 +58,10 @@ svg icons will not work, including system ones.
 
 At least Qt 6.6 is required.
 
+Developer builds with `-DBUILD_TESTING=ON` also require Qt Linguist Tools
+(`lrelease`) to compile translation test catalogs. This is not a runtime
+dependency or a requirement for builds with testing disabled.
+
 All features are enabled by default and some have their own dependencies.
 
 ### Crash Handler
--- a/changelog/next.md
+++ b/changelog/next.md
@@ -1,3 +1,10 @@
+## Features
+
+- Added shell translation catalogs in `i18n/qml_<language>.qm`, with runtime
+  language switching through `Qt.uiLanguage`. Place compiled Qt Linguist catalogs
+  beside the root QML file in `i18n/` and use `qsTr()` in translation bindings.
+  Reload the shell after updating catalogs.
+
 ## Bug Fixes
 
 - Fixed main process crashes on pam subprocess misbehavior.
--- a/shell.nix
+++ b/shell.nix
@@ -19,6 +19,7 @@ in pkgs.mkShell.override { stdenv = quickshell.stdenv; } {
     clang-tools
     parallel
     makeWrapper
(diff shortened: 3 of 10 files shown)


komagata:
@outfoxxed When you have a moment, could you let us know whether this approach to loading translation catalogs fits Quickshell's intended direction?

I'd like to move forward with Qt-based i18n in Omarchy, and this proposal would provide the foundation. Other shell developers have also expressed interest in the standard Qt workflow in #445, and all 30 Nix CI checks are currently passing.

Even a brief indication of the preferred API or design, ahead of a detailed review, would help us plan the next steps. Thanks!

outfoxxed:
I have no issue with the intention of adding translation support, but I need to look into the stability of translation contexts and if we should be using qtranslator here or something else.

komagata:
@outfoxxed Thanks for your reply! What alternatives to QTranslator are you considering?

komagata:
@outfoxxed Please let me know if there is anything I can do to help move this forward. I would be happy to investigate alternatives, test approaches, or help with anything else you need.

PR #1097, opened 2026-09-07
wayland/idle_notify: re-subscribe isIdle when the notification is replaced

### LINKED ISSUE #938: IdleMonitor goes permanently silent after a runtime write to `timeout`
Writing `timeout` on a live `Quickshell.Wayland.IdleMonitor` makes it go permanently silent: it behaves as if the existing `ext_idle_notification_v1` registration is torn down without a working replacement. The monitor neither keeps its old timeout nor adopts the new one — it never fires again until `enabled` is toggled (which does re-register correctly).

### Reproduction

Run this on a compositor with `ext-idle-notify-v1` and leave the session untouched (I ran it on a nested Hyprland that receives no input, so the session is idle from t=0):

```qml
import QtQuick
import Quickshell
import Quickshell.Wayland

ShellRoot {
    id: root

    property double t0: Date.now()
    function note(msg) { console.log("probe: +" + Math.round(Date.now() - root.t0) + "ms " + msg); }

    property bool moved: false

    // Proves idle notifications work at all during the run.
    property IdleMonitor control: IdleMonitor {
        timeout: 5
        onIsIdleChanged: root.note("control(5s fixed) isIdle=" + isIdle)
    }

    // If the write re-registered with the new value, this fires ~2s after the write.
    property IdleMonitor shortened: IdleMonitor {
        timeout: root.moved ? 2 : 3600
        onIsIdleChanged: root.note("shortened(3600->2 at t=3s) isIdle=" + isIdle)
    }

    // If the write were merely ignored (old registration kept), this fires at t=8s.
    property IdleMonitor lengthened: IdleMonitor {
        timeout: root.moved ? 999 : 8
        onIsIdleChanged: root.note("lengthened(8->999 at t=3s) isIdle=" + isIdle)
    }

    // Toggling `enabled` off and on *does* re-register: this re-fires ~2s after re-enable.
    property bool toggledOut: false
    property IdleMonitor toggled: IdleMonitor {
        enabled: !root.toggledOut
        timeout: 2
        onIsIdleChanged: root.note("toggled(2s) isIdle=" + isIdle)
    }

    property Timer move: Timer {
        interval: 3000; running: true
        onTriggered: { root.moved = true; root.note("wrote shortened.timeout 3600->2 and lengthened.timeout 8->999"); }
    }
    property Timer toggleOff: Timer {
        interval: 8000; running: true
        onTriggered: { root.toggledOut = true; root.note("toggled.enabled -> false"); }
    }
    property Timer toggleOn: Timer {
        interval: 8500; running: true
        onTriggered: { root.toggledOut = false; root.note("toggled.enabled -> true"); }
    }
    property Timer done: Timer {
        interval: 15000; running: true
        onTriggered: root.note("probe done")
    }

    Component.onCompleted: note("probe ready")
}
```

### Observed

```
probe: +0ms probe ready
probe: +2016ms toggled(2s) isIdle=true
probe: +3017ms wrote shortened.timeout 3600->2 and lengthened.timeout 8->999
probe: +5016ms control(5s fixed) isIdle=true
probe: +8016ms toggled(2s) isIdle=false
probe: +8016ms toggled.enabled -> false
probe: +8517ms toggled.enabled -> true
probe: +10517ms toggled(2s) isIdle=true
probe: +15016ms probe done
```

The two monitors whose `timeout` was written never fire:

- `shortened` (3600 → 2 at t=3s) is silent at t=5s and beyond, so the new value was not adopted.
- `lengthened` (8 → 999 at t=3s) is silent at t=8s and beyond, so the old registration was not kept either — it is gone.
- `control` firing at t=5s shows idle delivery works during the run, and `toggled` re-firing 2s after its `enabled` pulse shows toggling does rebuild the registration.

A longer differential run under a session lock showed the same signature: a monitor whose timeout was rebound 30 → 5 by the lock fired at neither 30s nor 5s over 50s, while a monitor with an unchanged timeout fired normally under the same lock.

### Expected

A `timeout` write on an enabled monitor re-registers with the new value (the protocol bakes the timeout into the notification at creation, so a replacement object is the only way) — or, failing that, at least keeps the old registration instead of losing both.

### Versions

- Quickshell 0.3.0 (Arch `quickshell` package). `src/wayland/idle_notify/monitor.{cpp,hpp}` are byte-identical between `v0.3.0` and current master (`28771c7c`), so this should reproduce on HEAD.
- Hyprland 0.56.1 (`ext-idle-notify-v1`), probe run on a nested instance.

`IdleMonitor::updateNotification()` in `src/wayland/idle_notify/monitor.cpp` looks like it should handle this — it is the change callback of `bParams`, deletes the old notification and creates a new one from the current params — so the surprising part is the asymmetry: an `enabled` toggle ends with a working registration and a `timeout` write ends with a dead monitor, though both funnel through the same callback. (I have not traced whether the replacement object is never created or is created but never propagates to `isIdle`.)

### Workaround

Pulse `enabled` off for one tick whenever the timeout changes; the re-enable re-registers with the new value already in place. That is what we ship: [`Services/System/IdleRung.qml`](https://github.com/danielbaldwin47/forest-shell/blob/main/Services/System/IdleRung.qml). Downstream tracking: danielbaldwin47/forest-shell#151.

voglster:
Fixes #938.

`IdleMonitor` goes permanently silent after a runtime write to `timeout`. The
notification is recreated correctly — the interesting part is that nobody is
listening to it.

## Cause

`updateNotification` frees the old notification and allocates the replacement
within the same call, so the allocator routinely returns the block just freed:

```
quickshell.wayland.idle_notify: Destroyed IdleNotification(0x7fda3454e210)
quickshell.wayland.idle_notify: Created   IdleNotification(0x7fda3454e210) with timeout: 20000
```

`bNotification` is only ever written once, at scope exit, and the new pointer
compares equal to the old one. `QObjectBindableProperty::setValue` returns
early on an equal value without calling `notify()`, so the `bIsIdle` binding
installed in `onPostReload` — which reads `bNotification` and then reaches
through it to that notification's own `bIsIdle` — is never re-evaluated and
never subscribes to the replacement. When the new notification receives
`idled`, nothing propagates.

The old `notification = nullptr` assigned the local variable, not the property,
so the property never observed an intermediate state.

This is why toggling `enabled` recovers a wedged monitor: that path is two
`updateNotification` calls with a real `nullptr` written to the property in
between.

Note the old code was only correct when the allocator happened to hand back a
different address. The fix removes the dependency on allocator behaviour rather
than relying on it.

## Fix

Clear `bNotification` before freeing the old notification, so the property
transitions `A -> nullptr -> B` and both writes notify. It also drops the
binding's subscription to the old notification while it is still alive.

## Tests

`src/wayland/idle_notify/test/monitor.cpp` adds two cases. A zero timeout is
specified to notify as soon as the seat is inactive, so neither has to wait out
a real idle period.

```
unfixed:  PASS freshMonitorReportsIdle
          FAIL liveTimeoutChangeKeepsIdleSubscription
fixed:    PASS both
```

`freshMonitorReportsIdle` is deliberately a real assertion rather than a guard,
so a regression in the construction path fails loudly instead of masking the
second test. Both skip when the compositor lacks `ext-idle-notify-v1`.

Full suite: 10/10 passing in 0.65s.

## The bug is allocator-dependent, and so is the test

Built with `-DASAN=ON`, the **unfixed** code passes both tests. ASAN's allocator
quarantines freed memory instead of handing it straight back, so the replacement
lands on a different address, `bNotification` genuinely changes, and the binding
re-subscribes. That is independent confirmation of the mechanism — remove the
address reuse and the bug disappears without touching `updateNotification`.

It also means this regression test cannot catch this bug under ASAN. No
black-box test can: without address reuse the old code is correct. Worth knowing
before concluding the test is ineffective if you run the suite sanitised.

ASAN reports nothing on either version, so this is a stale-value bug rather than
a memory-safety one. Clearing the property before the `delete` is defensive
ordering, not a use-after-free fix.

Two more things worth flagging, both happy to change:

- This is the first automated test under `src/wayland` and the first use of
  `QTRY_*` in the tree. It needs a live compositor, so it will skip in CI as
  currently configured — it is a regression test for developers rather than CI
  coverage.
- Because the control is now an assertion, a compositor that supports the
  protocol but mishandles zero timeouts will fail rather than skip.

---

Thanks for Quickshell — this is my first contribution here, so please tell me if
I've got the conventions wrong.

I came at this from the other end: my screen stopped blanking on Omarchy and I
went digging for why. It turned out to be this rather than anything in Omarchy's
shell, which is how I ended up in `idle_notify`.

Full disclosure on process: I used Claude Code to investigate this and write the
patch and tests. I directed the work, reviewed the diff line by line, and had it
walk me through the parts I didn't know — I can explain the reasoning behind
every line, and responsibility for it sits with me, which I understand is what
CONTRIBUTING asks for. Flagging it because I'd rather you know than wonder.

On style: I've run `clang-format`, followed the per-module `qs_test` pattern from
`src/core/test`, and matched the commit subject convention. But I write Python
day to day, so if the C++ idiom is off anywhere, say so and I'll fix it.


voglster:
I force pushed this to rebase on to latest master and fix the lint error i had on last build

PR #1106, opened 2026-09-09
wayland/lock: recover graphics initialization failures

acrogenesis:
Graphics initialization failure on a session-lock surface aborts Quickshell, leaving the session locked without an authentication UI.

Handle `sceneGraphError` and recreate the failed native window: Qt cannot retry RHI initialization on that window. Preserve the compositor lock, QML input state, geometry, visibility, color and focus. Retries back off from 1 to 30 seconds, reset after a rendered frame, transfer on reload, and stop after unlock or output removal.

Tested with Clang 22 and Qt 6.11.2:

- All-features Debug build and eight focused Qt tests pass across offscreen and opt-in Wayland runs.
- Dedicated Hyprland VM lock fixture: `EGL_BAD_ALLOC` injection, reload during pending retries, recovery, keyboard input and unlock pass with basic and threaded render loops.
- Full offscreen suite: 9/10 executables pass; existing `popupwindow::moveWithParent` failure remains.
- Formatting passes. Local lint reports only existing findings outside changed lines; custom tidyfox is unavailable.

Physical NVIDIA memory exhaustion, older Qt and multiple monitors remain untested. Persistent allocation failure still requires resources to become available.


outfoxxed:
Have you observed failure paths besides OOM on launch? I'd rather just trigger a wayland unlock if we fail to allocate all lock surfaces initially instead of doing an exponential backoff on vram.

If you're specifically observing a case where a scene graph error happens on unsuspend or something then that requires more consideration.

acrogenesis:
This happened when idle locking an already running desktop, rather than on resume:

The sequence of events was:
  1. vLLM occupied ~30 GB of VRAM, leaving roughly 388 MB free.
  2. The desktop reached its idle timeout and requested a lock.
  3. Qt hit EGL_BAD_ALLOC during graphics initialization and aborted Quickshell.
  4. Quickshell crashed seven times, exhausting the launcher's restart attempts. Hyprland remained alive.
  5. I recovered the shell using software OpenGL.

The backoff was my proposed recovery approach, rather than something required by an observed resume failure. Your suggestion to unlock if initial lock-surface allocation fails fits the confirmed case.

acrogenesis:
I implemented your idea over here https://github.com/quickshell-mirror/quickshell/pull/1175

PR #1128, opened 2026-09-13
x11/i3: build the module without requiring X11

qaqland:
BUILD.md documents i3/Sway as having no x11 dependency, but the module was only added from src/x11/CMakeLists.txt, so -DI3=ON -DX11=OFF silently lost Quickshell.I3.


outfoxxed:
thanks!

PR #1135, opened 2026-09-13
Fix/wired ethernet

Jaime23RL:
Fixes the bar never showing ethernet (reported in ilyamiro/serpantinum#165).

Root cause: noctalia-qs 0.0.12 removed Wired from DeviceType and ignored
NMDeviceType::Ethernet in nm/backend, so Quickshell.Networking never
exposed enp4s0 to QML. WifiWidget.qml isEthDevice() always false.

This restores Wired as in quickshell-mirror/quickshell:

- src/network/enums.hpp: add Wired=2, enums.cpp toString
- add src/network/wired.{hpp,cpp}: WiredDevice with hasLink/linkSpeed/network
- src/network/CMakeLists.txt: add wired.cpp
- src/network/nm/backend.cpp: handle Ethernet, create WiredDevice,
  hasLink via /sys/class/net/<iface>/carrier, linkSpeed via state/speed

Tested on CachyOS (B560M DS3H V2, MT7925, enp4s0 + wlan0):
- before: qmltypes ["None","Wifi"], DeviceType.Wired undefined,
  Networking.devices 1, log Ignoring unsupported device
- after: qmltypes ["None","Wifi","Wired"], DeviceType.Wired=2,
  devices 2, WifiWidget shows eth


outfoxxed:
send this to noctalia

PR #1166, opened 2026-09-20
service/tray: forward Wayland activation tokens on primary clicks

domenkozar:
Tray clicks currently send only `Activate(0, 0)`. A native Wayland application cannot use the panel's input event to request focus, so compositors such as niri may leave its existing window unfocused.

Request a token through `QWaylandWindow::requestXdgActivationToken()` using Qt's last input serial and window, then deliver it with `ProvideXdgActivationToken` before calling `Activate`. Implement the corresponding layer-shell override so this works for panels as well as ordinary Qt windows. The activation protocol implementation belongs to layer-shell, and the tray service uses Qt's window API.

The existing `SystemTrayItem.activate()` QML API is unchanged. Missing input context, unavailable activation support, a one-second timeout, and applications that reject the optional token method all preserve ordinary activation. Ignore callbacks after the tray item disappears. Allow only one pending tray token request per window, so a late response cannot be reused for another tray click; overlapping clicks use ordinary activation.

Validation:

- Full debug build with precompiled headers disabled and tests enabled; all 9 tests pass with `QT_QPA_PLATFORM=wayland`. The popup-positioning test fails with Qt's offscreen/minimal backends.
- Formatting and clang-tidy with the repository's tidyfox plugin pass for the changed code.
- Tray module builds with Wayland disabled; configuration also succeeds with layer-shell disabled.
- Temporary lifecycle harness checks token delivery, single completion, timeout, late responses, tray context and window destruction, and a callback that destroys its window.
- Live niri layer-shell pointer clicks: both token-aware and legacy mock tray items receive a nonempty token followed by `Activate`; the legacy item rejects the unsupported token method and still activates.

Prepared with AI assistance for human review.


outfoxxed:
Do you have a reason to avoid QWaylandWindow::requestXdgActivationToken?

domenkozar:
I used the protocol directly because Qt’s method needs support from Quickshell’s layer-shell implementation, which currently just emits an empty token. 

We could add that support and then use QWaylandWindow::requestXdgActivationToken() in the tray code.

outfoxxed:
> We could add that support and then use QWaylandWindow::requestXdgActivationToken() in the tray code.

Please do

domenkozar:
CI failed unrelatedly

